PR #11, opened 2022-04-14
Update pytest.yml

mario-valdivia:


mario-valdivia:
added build docs in auto tests

PR #12, opened 2022-04-20
Dev/0.3.1

jsenellart:
- fix `pdisplay` on int/float/complex
- improve efficiency of building unitaries
- optimize call to SLOS
- extend `random_unitary` with provided random numbers
- improve circuit rendering
- improve triangle decomposition: possibility to pass 1 parameter unitary block, pass constraints on parameters to avoid phases when not necessary, values of parameters bound in the search, introduce identification of permutations
- improve generic_interferometer implementation - new parameter to add phase layer and depth can be used to generate depth > m (see https://arxiv.org/pdf/1910.13001)
- improve polarization encoding to path encoding mapping - by allowing 2 orthogonal phases on each input modes 
- check parameter boundaries 
- more unit tests


jsenellart:
Hello @mario-valdivia - I opened this PR - tests fail because they need updated version of quandelibc - I let you handle?

PR #23, opened 2022-05-07
Add use case MZI with imperfect SPS

MathiasPnt:
Use case to compute the output state probability distribution or the outcome (click detectors) of a Mach-Zehnder interferometer with an imperfect quantum-dot single-photon source. The simulation is in good agreement with the model developed by Ollivier _et al._ in https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.126.063602.


jsenellart:
Hello @MathiasPnt - thanks for your contribution ! To make this more visible, I propose that we make it available as a notebook. If you can submit a first version, we will glad to polish it before publication. We are targeting 0.5.0 within 2 weeks, it would be great to have it integrated by then !

PR #35, opened 2022-06-13
Update probability_distribution source.py

MathiasPnt:
Main updates:
- update the constructor of the class Source with input parameters that can be measured experimentally.
- update probability_distribution so that: the noise photon is labelled 1 if "distinguishable", 0 otherwise. The probability of having 2 photon p2 is defined by g2 = 2p2/(p1+2p2)**2 & p1+p2=beta. 

Others:
- update the test for source.py and processor.py
- Update pcvl_ImperfectSource.py. The current version will be implemented in a notebook.


jsenellart (commented):
Hello @MathiasPnt, for back-compatibility I propose that we keep purity as a legacy parameter showing a warning when used, but not breaking potential user code already using it.

We would set `multiphoton_component=1-purity` - result would not be exactly the same, but it would still run. let me know if ok for you.

MathiasPnt:
@jsenellart yes that would be fine. In the definition we should then define purity as the single-photon purity (P = 1-g2) and not p1 as it is now defined in the main branch.

perceval/components/source.py
```diff
@@ -28,58 +28,93 @@
...
+        g2 = self.multiphoton_component
+        mu = self.brightness
+        eta = self.overall_transmission
+
+        beta = mu - g2 * mu ** 2 / 2
+        p2 = g2 * mu ** 2 / 2
+        p1 = beta - p2
+
         svd = SVDistribution()
-        if self.brightness != 1:
```
jsenellart, line 60: did you remove this on purpose? without this, the Source is not generating `[0]` states and tests in processor with several source fail

jsenellart (changes requested):

perceval/components/source.py
MathiasPnt: Yes! Because i) we will always postselect on detected events and ii) the losses and the brightness of the sources are already taken into accounts with mu and eta.

PR #54, opened 2022-07-15
Dev/connectors

jsenellart:
first integration of qiskit converter with generic component library available in `library.catalog`. converter is now returning a processor which source are set on `|0>` for each qbits. Heralding logic is hardcoded and needs to be made generic.


ericbrts:
All the code was included in release 0.6.0 via another PR, so I close this one.

PR #77, opened 2022-08-30
Pcvl 169 add utility functions to state vectors

jsenellart:


perceval/utils/statevector.py
```diff
@@ -470,6 +472,61 @@ def __sub__(self, other):
...
+    def sample(self) -> AnnotatedBasicState:
+        r"""Sample a single state from the statevector.
+        It does not perform a measure - so do not change the value of the statevector
+
+        :return: an annotated basic state
+        """
+        p = random.random()
+        idx = 0
+        keys = list(self.keys())
+        self._normalize()
```
ericbrts, line 484: Shouldn't _normalize() be called when you change the state vector (i.e. in __add__, __mul__, __sub__) rather than when you use it?
This means there are times when the state vector is not normalized but can still be used (or at least displayed).

jsenellart: by default, we want the normalization to be lazy since we can still make operations on it before requiring to have it normalized. However, this means that each external access (display, or even accessor to coefficients must catch the call and call normalization) - and each internal access as in this case should be careful of normalizing it. I keep as is for the moment, but will make a specific review on that.

PR #86, opened 2022-09-14
PCVL-139: Direct access to components in a circuit

Elian-OR:


ericbrts:
Tested with Autotest manual run, on my fork: https://github.com/ericbrts/Perceval/actions/runs/3059021726

PR #101, opened 2022-10-04
PCVL-217 remove SF backend

ericbrts:
@jsenellart : I've removed all reference to Strawberry Fields from the bibliography, is that alright?


jsenellart (approved):
yes it is.

PR #102, opened 2022-10-05
PCVL-226 remote processor + job progress cb

mario-valdivia:


perceval/runtime/remote_processor.py
```diff
@@ -0,0 +1,143 @@
...
+    for v in specs.values():
+        if 'name' in v:
+            return v['name']
+    return None
+
+
+def _split_platform_and_backend_name(name: str):
+    backend_name = 'Naive'
+    platform_name = name
+    colon_pos = name.find(":")
```
ericbrts, line 52: For readability, I'd suggest:
```
split_name = name.split(':')
if len(split_name) == 2:
    platform_name = split_name[0]
    backend_name = split_name[1]
```

perceval/runtime/remote_backend.py
```diff
@@ -28,118 +28,89 @@
...
-    for method in dir(cls):
-        if method.startswith('async_'):
-            sync_name = method.removeprefix('async_')
-            setattr(cls, sync_name, _sync_wrapper(cls, method))
-    return cls
 
 
 ################################
 @generate_sync_methods
 class RemoteBackend(Backend):
```
ericbrts, line 43: The payloads we need to generate do not follow the Backend interface. As such, I thought we said the RemoteBackend class should disappear and its responsibility (be able to generate payloads) transferred to the RemoteProcessor (which actually follows the API we need to expose - samples(), sample_count() and probs())

perceval/runtime/remote_processor.py
```diff
@@ -0,0 +1,143 @@
...
+    def type(self) -> ProcessorType:
+        return self._type
+
+    def with_input(self, input_state: BasicState) -> None:
+        self._input_state = input_state
+
+    @property
+    def available_sampling_method(self) -> str:
+        for k, v in _extract_commands(self._specs):
+            if v.startswith('sample'):
```
ericbrts, line 112: I'm not sure this'll work as "probs" should be a valid answer in some cases. We should discuss this.

perceval/runtime/remote_processor.py
```diff
@@ -0,0 +1,143 @@
...
+
+        return self._backend.async_probs(self._input_state)
+
+    def get_circuit_parameters(self) -> Dict[str, Parameter]:
+        pass
+
+    def set_circuit_parameters(self, params: Dict[str, Parameter]) -> None:
+        pass
+
+    def resume_job(self, job_id:str, deserializer: Callable = None):
```
ericbrts, line 140: How should the user know which deserializer to use when calling `resume_job`?

requirements.txt
```diff
@@ -4,3 +4,11 @@ tabulate
 pytest
 setuptools
 multipledispatch
+
```
ericbrts, line 7: protobuf is missing

perceval/runtime/remote_backend.py
ericbrts: Also, the payload should contain Processor._parameters key/value pairs, so it makes even more sense to give the responsibility of creating payloads to the RemoteProcessor.

For instance, the noise parameter, expected by some simulators, is meant to be in Processor._parameters dict.

PR #145, opened 2022-11-10
PCVL-268 conversion layer calls undefined function

ericbrts:


perceval/components/processor.py
```diff
@@ -445,10 +445,7 @@ def sample_count(self, count: int, progress_callback: Callable = None) -> Dict:
         raise RuntimeError(f"Cannot call sample_count(). Available method are {self.available_commands}")
 
     def _sample_inputs(self, count, non_null=False) -> List[StateVector]:
-        sampled = self._inputs_map.sample(count, non_null=non_null)
-        if count == 1:
-            return [sampled]
-        return sampled
+        return self._inputs_map.sample(count, non_null=non_null)
```
Alex-PLACET, line 448: Does this function still useful as it just call a function with same parameters wiuthout any logic ?

ericbrts: I guess not: I'm removing it right now.

PR #155, opened 2022-12-12
Correct probability_distribution of class Source

MathiasPnt:
probability_distribution of class Source for v0.7.3. The class Source now includes:
- occupation factor
- multiphoton component
- overall_transmisison = losses * brightness

### DIFF
--- a/perceval/components/source.py
+++ b/perceval/components/source.py
@@ -57,14 +57,6 @@ class Source:
         else:
             assert purity is None, "cannot set both purity and multiphoton_component"
             
-        if occupation_factor is None:
-            if brightness is None:
-                ocupation_factor = 1
-        else:
-            occupation_factor = brightness
-        else:
-            assert brightness is None, "cannot set both brightness and occupation_factor"
-            
         self.brightness = brightness
         self.occupation_factor = occupation_factor
         self.overall_transmission = overall_transmission
--- a/perceval/components/source.py
+++ b/perceval/components/source.py
@@ -21,32 +21,56 @@
 # SOFTWARE.
 
 import math
+import numpy as np
 
 from perceval.utils import SVDistribution, StateVector
-from typing import Dict
+from typing import Dict, Literal
 
 
 class Source:
-    r"""Definition of a single photon source. The default source is a perfect source (emits exactly 1 indistinguishable
-    photon each period).
-
-    :param brightness: the brightness of the source defined as the percentage of unique photon generation
-    :param purity: the ratio of time when photon is emitted alone
-    :param purity_model: `"random"` if additional photons are distinguishable, `"indistinguishable"` otherwise
-    :param indistinguishability: indistinguishability parameter as defined by `indistinguishability_model`
-    :param indistinguishability_model: `"homv"` defines indistinguishability as HOM visibility, `"linear"` defines
-        indistinguishability as ratio of indistinguishable photons
-    :param context: gives a local context for source specific features, like `"discernability_tag"`
-    """
     def __init__(self, brightness: float = 1,
-                 purity: float = 1,
-                 purity_model: str = "random",  # Literal["random", "indistinguishable"]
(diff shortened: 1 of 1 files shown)


perceval/components/source.py
```diff
@@ -21,66 +21,123 @@
...
-            svd[StateVector([1], {0: ["_:%s" % random_feat]})] = distinguishability*self.brightness*self.purity
-            svd[StateVector([1], {0: ["_:0"]})] = (1-distinguishability)*self.brightness*self.purity
+                if self._multiphoton_model == "distinguishable":
+                    svd[StateVector([2], {1: {"_": 0}, 2: {"_": noise_photon}})] = eta**2*p2
+                else:
+                    svd[StateVector([2])] = eta**2*p2
+
+        if distinguishability != 0:
+            if self._multiphoton_model == "distinguishable":
+                svd[StateVector([1], {1: {"_": distinguishable_photon}})] = eta*distinguishability*p1+eta*(1-eta)*distinguishability*p2
```
Alex-PLACET, line 126: l.126 and l.130 are the same. You can put it outside the condition scope

perceval/components/source.py
```diff
@@ -21,66 +21,123 @@
...
+                multiphoton_component = 0
+            else:
+                multiphoton_component = 1 - purity
+        else:
+            assert purity is None, "cannot set both purity and multiphoton_component"
+            
+        if occupation_factor is None:
+            if brightness is None:
+                ocupation_factor = 1
+        else:
```
Alex-PLACET, line 63: Warnin: You have an issue in the indentation. Here the else if for l.60

perceval/components/source.py
```diff
@@ -21,66 +21,123 @@
...
-        r""":return: SVDistribution on 1 mode associated to the source
+    def probability_distribution(self):
+        r"""returns SVDistribution on 1 mode associated to the source
         """
+
+        g2 = self.multiphoton_component
+        eta = self.overall_transmission*self.brightness/self.occupation_factor
+        px = self.occupation_factor
+
+        # g2 = 2p2/(p1+2p2)**2
```
Alex-PLACET, line 89: Do you need to keep these commented lines ?

perceval/components/source.py
```diff
@@ -21,66 +21,123 @@
...
         self.indistinguishability = indistinguishability
         self._indistinguishability_model = indistinguishability_model
         assert self._indistinguishability_model in ["homv", "linear"], "invalid value for indistinguishability_model"
         self._context = context or {}
         if "discernability_tag" not in self._context:
             self._context["discernability_tag"] = 1
 
-    def probability_distribution(self) -> SVDistribution:
-        r""":return: SVDistribution on 1 mode associated to the source
+    def probability_distribution(self):
```
Alex-PLACET, line 81: This function is a bit big, can you split it in the conditions ?

Alex-PLACET:
Can you add some unit test ?

perceval/components/source.py
Aubaert: There is also a missing c in occupation factor

MathiasPnt: This lines were unnecessary. I found a fix using "eta = self.overall_transmission*self.brightness/self.occupation_factor". I removed them and set occupation_factor=1 and brightness=1 by default.

perceval/components/source.py
MathiasPnt: This helps understand the physics behind the code. Not necessary but helpful.

perceval/components/source.py
MathiasPnt: self._multiphoton_model can be either "indistinguishable" or "distinguishable"

perceval/components/source.py
Alex-PLACET: So please add a comment that explain that is helpul to understand the code.

PR #156, opened 2022-12-13
PCVL-246 customizable job name

ericbrts:
Job names can be customized through algorithm or job instances.

1. An algorithm has a `default_job_name` field it uses to name the jobs it creates. If `default_job_name` is missing, it must generate a name on its own (sampler names the jobs with the primitive that has been called - probs, samples or sample_count).

2. Jobs have a name setter:
```python
job.name = "My job name!"
```


perceval/runtime/remote_job.py
```diff
@@ -97,8 +116,12 @@ def execute_async(self, *args, **kwargs):
                 if self._job_context is None:
                     self._job_context = {}
                 self._job_context["mapping_delta_parameters"] = self._delta_parameters
+            kwargs = self._handle_unnamed_params(args, kwargs)
             kwargs['job_context'] = self._job_context
-            self._id = self._fn(*args, **kwargs)
+            self._payload['job_name'] = self._name
+            self._payload['payload'].update(kwargs)
```
Aubaert, line 122: It supposes that the payload has been created by a processor before (maybe add a docstring ?)

Aubaert (changes requested):

perceval/algorithm/analyzer.py
```diff
@@ -104,7 +104,9 @@ def compute(self, normalize=False, expected=None, progress_callback=None):
         # Compute probabilities for all input states
         for idx, i_state in enumerate(self.input_states_list):
             self._processor.with_input(i_state)
-            probs_output = self._sampler.probs()
+            job = self._sampler.probs
+            job.name = self.default_job_name or f'analyzer ({idx+1}/{len(self.input_states_list)})'
```
mario-valdivia, line 108: def analyser.__init():
  self.default_name = "analyzer"

def analyzer.compute():
   job.name = f'{self.default_job_name} ({idx+1}/{len(self.input_states_list)})'

mario-valdivia (changes requested):

Alex-PLACET:
Can you squash the commits ?

perceval/algorithm/sampler.py
```diff
@@ -74,9 +75,11 @@ def _generic(self, method: str):
             job_context = None
             if converter:
                 job_context = {"result_mapping": ['perceval.utils', converter.__name__]}
-            rj = RemoteJob(getattr(self._processor, "async_"+primitive),
-                           self._processor.get_rpc_handler(), delta_parameters=delta_parameters,
-                           job_context=job_context)
+            payload = self._processor.prepare_job_payload(primitive)
+            job_name = self.default_job_name if self.default_job_name is not None else method
```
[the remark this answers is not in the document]
ericbrts: The thing is, the command of a Sampler is not the correct one (i.e. the one the user asked for) whenever a post-conversion is needed. So I prefer setting `method` dynamically when the `default_job_name` was not set by the user.

PR #157, opened 2022-12-14
PCVL-306 small improvements

ericbrts:


perceval/backends/cliffords2017.py
```diff
@@ -45,7 +45,9 @@ def prob_be(self, input_state, output_state, n=None, output_idx=None):
 
     def sample(self, input_state):
```
Alex-PLACET, line 46: Can you add typing in the signature ?

perceval/components/abstract_processor.py
```diff
@@ -90,6 +94,10 @@ def parameters(self):
     def clear_parameters(self):
         self._parameters = {}
 
+    def clear_input_and_circuit(self):
```
Alex-PLACET, line 97: In which workflow do we have to use these clear_XXX methods ?
Is it just to avoid to create a new processor object ?

perceval/components/abstract_processor.py
```diff
@@ -47,21 +47,25 @@ def __init__(self):
...
-        self._in_ports: Dict = {}
-        self._out_ports: Dict = {}
         self._thresholded_output: bool = False
         # Mode post selection: expect at least # modes with photons in output
         self._min_mode_post_select = None
-        self._post_select = None
+
+        self._reset_circuit()
+
+    def _reset_circuit(self):
```
Alex-PLACET, line 56: IMO, the variables which are the "Circuit" should be in a dedicated class

perceval/components/abstract_processor.py
ericbrts: Yes, `clear_input_and_circuit` allows to keep all the `Processor` config and refresh only circuit and input state. It is required to improve the syntax of upcoming code.

`clear_postprocess()` can already be useful during Processor composition with others, as a postprocess function blocks any further composition.

perceval/backends/cliffords2017.py
ericbrts: OK

perceval/components/abstract_processor.py
ericbrts: Let's discuss this during a specific meeting, as this is too big a request to be implemented in this branch.

perceval/components/abstract_processor.py
Alex-PLACET: It looks strange to have a class named Processor where you can clear the circuit.
Also I prefer to create a new object and copy several parameters than clearing some members of an existing one.

Alex-PLACET:
Can you squash the commits ?

perceval/backends/cliffords2017.py
```diff
@@ -43,9 +44,11 @@ class CliffordClifford2017Backend(Backend):
     def prob_be(self, input_state, output_state, n=None, output_idx=None):
         raise NotImplementedError(f'Cannot use prob_be on {self.name}')
 
-    def sample(self, input_state):
+    def sample(self, input_state: Union[BasicState, StateVector]) -> BasicState:
         if isinstance(input_state, StateVector):
-            input_state = input_state.sample()
+            if len(input_state) != 1:
+                raise RuntimeError(f"{self.name} cannot sample with a mixed state input")
```
Aubaert, line 50: It's not mixed state input (= SVDistribution) that lead to a problem, but superposed pure states input (= StateVector of len > 1). The assertion is right, not the message

ericbrts: Right, thanks. Fixed!

perceval/components/abstract_processor.py
Aubaert: In case we have an optimisation process that tries to place components on its own, it can be useful to clear the processor as the number of modes or the input state can change during the optimisation process, as well as the circuit itself.
If you want, you can of course redefine at each iterations the "constants" of your processor, but you can also only change these parameters. This avoids for example to ask the cloud at each iteration the specs of a RemoteProcessor and deserialise them (which is done automatically).

PR #158, opened 2022-12-16
Factorised source creation

Aubaert:


ericbrts (approved):

perceval/components/source.py
```diff
@@ -69,81 +71,65 @@ def __init__(self, brightness: float = 1,
         assert self._indistinguishability_model in ["homv", "linear"], "invalid value for indistinguishability_model"
         self._context = context or {}
         if "discernability_tag" not in self._context:
-            self._context["discernability_tag"] = 1
+            self._context["discernability_tag"] = 0
```
Alex-PLACET, line 74: Was it a bug ?
I don't see unit test modification.

perceval/components/source.py
```diff
@@ -69,81 +71,65 @@ def __init__(self, brightness: float = 1,
...
 
-    def probability_distribution(self):
-        r"""returns SVDistribution on 1 mode associated to the source
-        """
+    def get_tag(self, tag, add=False):
```
Alex-PLACET, line 76: Please add a test

perceval/components/source.py
```diff
@@ -69,81 +71,65 @@ def __init__(self, brightness: float = 1,
...
+        if add:
+            self._context[tag] += 1
+        return self._context[tag]
 
+    def _get_probs(self):
```
Alex-PLACET, line 81: Please add a unit test

Alex-PLACET (changes requested):
The code is far better and readable but you have to add several unit tests.

perceval/components/source.py
Aubaert: It is now incremented within the get_tag function, before being given instead of after, so it doesn't change anything

Alex-PLACET (approved):

PR #160, opened 2022-12-19
PCVL-230 PCVL-308 issue fixing

ericbrts:


perceval/backends/template.py
```diff
@@ -213,8 +219,10 @@ def probampli(self,
         :param n:
         :return: complex probability amplitude
         """
+        self._check_state_size(input_state)
+        self._check_state_size(output_state)
         if input_state.n == 0:
-            return output_state.n == 0
+            return complex(1) if output_state.n == 0 else complex(0)
```
Aubaert, line 225: I find writing 1 + 0j more readable but its a personal opinion.

tests/test_simulators.py
```diff
@@ -64,40 +64,62 @@ def check_output(simulator, input_state, expected):
...
         check_output(s, pcvl.BasicState([0, 0]), {pcvl.BasicState("|0,0>"): 1})
         check_output(s, pcvl.BasicState([0, 1]), {pcvl.BasicState("|0,1>"): 1})
         check_output(s, pcvl.BasicState([1, 1]), {pcvl.BasicState("|1,1>"): 1})
 
 
-def test_sim_indistinct_sym():
+def test_simulator_wrong_size():
+    circuit = pcvl.Circuit(2)
+    state = pcvl.BasicState([1, 1, 1])
     for backend in ["SLOS", "Naive"]:
```
Aubaert, line 89: Add MPS backend

docs/source/components.rst
```diff
@@ -106,7 +106,7 @@ To create a beam splitter object with a given reflectivity value:
 >>> beam_splitter = BS(BS.r_to_theta(R))
 
 By default
-``theta`` is :math:`\pi/4`,
+``theta`` is :math:`\pi/2`,
```
Alex-PLACET, line 109: Should we consider that important enough to warn the user ?

Aubaert: I think it was a doc mistake, so it's not mandatory.

Aubaert (approved):

docs/source/components.rst
ericbrts: It 's a doc mistake. I fixed the default value above in the same page, and didn't check if the value further.. but the value was repeated elsewhere (here)!

PR #164, opened 2022-12-21
Build and deploy docs correction

julienCALISTO:


ericbrts:
We used only part of the changes which were directly commited in release/0.7.3 branch.
So this PR is closed without being merged

PR #171, opened 2023-01-30
PCVL-315 circuit optimizer decomposition

ericbrts:


perceval/utils/algorithms/circuit_optimizer.py
```diff
@@ -0,0 +1,84 @@
...
+        self._threshold = 1e-6
+        self._trials = 4
+
+    @property
+    def threshold(self):
+        return self._threshold
+
+    @threshold.setter
+    def threshold(self, value):
+        if value < 0 or value > 1:
```
Alex-PLACET, line 43: `not (0 < value < 1)` is most expressive IMO

PR #172, opened 2023-01-30
fix missing parameter in RemoteJob initializer

jsenellart:
fix missing parameter in RemoteJob initializer


ericbrts:
PR closed because the issue was fixed on 0.8.0 branch

PR #176, opened 2023-02-03
PCVL-261 identity block param decomposition

ericbrts:


perceval/utils/algorithms/decomposition.py
```diff
@@ -59,7 +59,8 @@ def decompose_triangle(u,
                        permutation,
```
Alex-PLACET, line 59: This function is soo big. Is it possible to split it ?

PR #179, opened 2023-02-07
PCVL-316 qiskit code reference generation error

ericbrts:


jsenellart (approved):
looks totally right to me!

Issue #189, opened 2023-02-27
drawSvg 2.0 has a new api, causes import failure


abdullahkhalids:
**Prerequisite**
Please ensure that this bug has not been answered on the [Perceval forum](https://perceval.quandela.net/forum) Done

**Describe the bug**
drawsvg (previously drawSvg) has changed its [naming scheme](https://pypi.org/project/drawsvg/) from CamelCase to underscore based, in the 2.0 major release.

Installing drawsvg with version greater than 2.0 causes `import perceval` to fail, because of `import drawSvg` statements in several places. 

**To Reproduce**
1. On a clean system, `python -m pip install perceval-quandela`
2. On the python console `import perceval`
3.  Error relating to `import drawSvg` pops up

**Expected behavior**
Import should occur with no errors.

**Desktop (please complete the following information):**
 - OS: Debian
 - CPU Arch: x86_64
 - Python Version: 3.10
 - Perceval Version: 0.7.3

**Additional context**
The temporary fix is to add `drawsvg<2.0` in the requirements.txt. Long term, perceval should start using the latest version of drawsvg. As many of drawsvg function names have changed, this will require some work.

ericbrts:
Hello,

Thank you for your detailed message.
You're right, drawsvg released a major update last Sunday, breaking its own API.
Perceval will support drawsvg 2.0+ starting from the upcoming 0.8.0 version.

For now, as you proposed, users willing to install new environments with Perceval 0.7.x will have to manually downgrade drawsvg.
This can be done, in command line, as easily as:
`(venv) $ pip install drawsvg<2.0`

Best regards,
Eric

abdullahkhalids:
Thank you. Small correction: one needs quotes to prevent the < be interpreted as a shell redirect operator.
```
pip install "drawsvg<2.0"
```

ericbrts:
Hi,

Perceval 0.8.0 was released last Friday and uses drawsvg 2, fixing this issue for good.

Thanks again for your workaround.
Best regards,
Eric

PR #192, opened 2023-03-08
drawsvg<2.0 is a now a requirement

mbassiouny33:
Hi,

In https://pypi.org/project/drawsvg/#description  under section "Upgrading from version 1.x"

The doc authors mention few changes regarding capital "S". It seems Quandela was written for drawSvg 1.x and now when I try to import Perceval it fails as the version is not specified in your req. and therefore pip automatically installs the latest version.

I can think of Two solutions:
1- Adapt all code using DrawSvg to the new version [proper solution imho].
2- Force the installed version to the best know version that you have tested (1.9 seems to work fine for me)[quick workaround].

This PR implement the quick n dirty workaround, I didn't dare to go for the 1st option as I am new to Perceval 😅...

Best

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,5 +11,5 @@ scipy
 tqdm
 Deprecated
 requests
-drawSvg
+drawSvg<2.0
 protobuf>=4.21.2
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setuptools.setup(
     ],
     packages=package_list,
     install_requires=['sympy', 'numpy', 'scipy', 'tabulate', 'matplotlib', 'quandelibc~=0.6.0', 'multipledispatch',
-                      'protobuf>=4.21.2', 'drawSvg', 'Deprecated', 'requests'],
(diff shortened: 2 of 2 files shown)


ericbrts:
Hi,

Thanks for your pull request.

You're right, Perceval 0.7.* only works with drawsvg 1.*.
The fix to use drawsvg 2.0 is already done in the 0.8 branch and waiting to be released as soon as possible and a manual (but easy) work-around exists (https://github.com/Quandela/Perceval/issues/189). Thus, we wouldn't like to integrate your changes in requirements.txt and setup.py, just before the 0.8.0 release.

About the typo in the documentation, good catch! However, the picture is not matching the matrix, nor the code either.
The picture corresponds to `c = comp.PERM([2, 3, 1, 0])`. Could you update the documentation with this line of code, and change the latex matrix accordingly, please? This way, we won't have to update the picture.
If you can't, don't worry, I can handle it!

Best regards,
Eric

mbassiouny33:
@ericbrts done!

code = matrix= picture now.

bien vu pour le code aussi !

And no worries for the the drawsvg version issue as long as you are already aware and have a fix planned for the next release.

PR #213, opened 2023-06-06
Pcvl 436 compress feature in serialization

raksharuia:


perceval/serialization/serialize.py
```diff
@@ -39,74 +41,103 @@
...
+
+@to_compress
 @dispatch(object)
-def serialize(obj) -> object:
+def serialize(obj, compress=False) -> object:
     return obj
 
 
-def serialize_to_file(obj, filepath: str) -> None:
+def serialize_to_file(obj, filepath: str, compress=False) -> None:
```
ericbrts, line 140: You're not using `compress` parameter in the function. I think it should be passed to `serialize(obj, compress=compress)` next line.

perceval/serialization/serialize.py
```diff
@@ -39,74 +41,103 @@
 import json
 
 
+def to_compress(func):
+    @wraps(func)
+    def compressor(*args, **kwargs):
+        if kwargs.get('compress', False):
+            serialized_string = func(*args, **kwargs)  # serialized circuit : string format
```
ericbrts, line 48: Could you remove the reference to the "circuit" in this line comment?

ericbrts (changes requested):

PR #219, opened 2023-06-14
Feature/control z

lukamusic:
Added a CZ gate to the catalog, based on the design of https://arxiv.org/abs/quant-ph/0110144.

### DIFF
--- a/perceval/components/core_catalog/heralded_cz.py
+++ b/perceval/components/core_catalog/heralded_cz.py
@@ -36,7 +36,7 @@ ctrl (dual rail) ─────┤     ├───── ctrl (dual rail)
 data (dual rail) ─────┤     ├───── data (dual rail)
                  ─────┤     ├─────
                       ╰─────╯"""
-
+	
     theta_1 = 2*np.pi*54.74/180
     theta_2 = 2*np.pi*17.63/180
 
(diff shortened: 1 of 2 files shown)


perceval/components/core_catalog/heralded_cz.py
```diff
@@ -0,0 +1,71 @@
...
+            .add((1, 2), PERM([1, 0]))
+            .add((2, 3, 4, 5), last_modes_cz)
+            .add((1, 2), PERM([1, 0])))
+
+        if self._opt('type') == AsType.CIRCUIT:
+            return c_hcz
+        elif self._opt('type') == AsType.PROCESSOR:
+            p = Processor(self._opt('backend'), c_hcz)
+            return p.add_port(0, Port(Encoding.DUAL_RAIL, 'ctrl')) \
+                    .add_port(2, Port(Encoding.DUAL_RAIL, 'data')) \
```
ericbrts, line 69: If I understand correctly, a CZ gate ports are symmetrical. Maybe we should use a better naming than "ctrl" and "data". What do you think?
If this changes, it should also be changed in the inline documentation at the top of the file.

perceval/components/core_catalog/heralded_cz.py
```diff
@@ -0,0 +1,71 @@
...
+            .add((1, 2), PERM([1, 0]))
+            .add((0, 1), BS.H(theta=self.theta1))
+            .add((2, 3), BS.H(theta=self.theta1))
+            .add((1, 2), PERM([1, 0]))
+            .add((0, 1), BS.H(theta=-self.theta1))
+            .add((2, 3), BS.H(theta=self.theta2)))
+
+        c_hcz=(Circuit(6, name="Heralded CZ")
+            .add((1, 2), PERM([1, 0]))
+            .add((2, 3, 4, 5), last_modes_cz)
```
ericbrts, line 61: Could you add `merge=True` as a parameter of the `add` call? It will "flatten" the components inside the gate and improve how it's `pdisplay`ed

lukamusic: Done

perceval/components/core_catalog/heralded_cz.py
```diff
@@ -0,0 +1,71 @@
...
+    str_repr = r"""                      ╭─────╮
+ctrl (dual rail) ─────┤     ├───── ctrl (dual rail)
+                 ─────┤     ├─────
+                      │     │
+data (dual rail) ─────┤     ├───── data (dual rail)
+                 ─────┤     ├─────
+                      ╰─────╯"""
+	
+    theta_1 = 2*np.pi*54.74/180
+    theta_2 = 2*np.pi*17.63/180
```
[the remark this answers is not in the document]
lukamusic: Done

perceval/components/core_catalog/heralded_cz.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __init__(self):
+        super().__init__("heralded cz")
+        self._default_opts['type'] = AsType.PROCESSOR
+
+    def build(self):
+	#the matrix of this first circuit is the same as the one presented in the reference paper, the difference in the second phase shift - placed on mode 3 instead of mode 1 - is due to a different convention for the beam-splitters (signs inverted in second column).
```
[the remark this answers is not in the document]
lukamusic: Done

perceval/components/core_catalog/heralded_cz.py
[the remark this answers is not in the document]
lukamusic: The additional 2 factor takes into account the difference between the beam-splitter conventions (Perceval vs the paper), I've added a comment to clarify

perceval/components/core_catalog/heralded_cz.py
lukamusic: I think we still want to have the terms control and data/target since that's the way that we usually talk about any controlled gate. I didn't see any other terminology used in this scenario.

PR #240, opened 2023-07-18
PCVL-482 Fix noisy algorithm (Simulator.probs_svd) for superposed states

ericbrts:


SwgAnno:
Test Comment

SwgAnno (approved):
The algorithm seems good to me now. I can understand how the injecting and merging of annotated output states that you had to add make the whole thing much slower.

PR #242, opened 2023-07-20
Pcvl 484 fix exponentiation

MarionQuandela:


tests/test_qmath.py
```diff
@@ -0,0 +1,45 @@
...
+    assert exponentiation_by_squaring(8, 2), 64
+    assert exponentiation_by_squaring(52, 13), 20325604337285010030592
+    bs = pcvl.BasicState("|1,0,1,0,0>")
+    assert bs**3, bs*bs*bs
+    sv = pcvl.StateVector(pcvl.BasicState("|1,0,1,0,0>")) + pcvl.StateVector(pcvl.BasicState("|1,0,0,1,0>"))
+    assert sv**2, sv*sv
+    assert sv**5, sv*sv*sv*sv*sv
+    sv = pcvl.StateVector(pcvl.BasicState('|1,0,4,0,0,2,0,1>')) - pcvl.StateVector(pcvl.BasicState('|1,0,3,0,2,1,0,1>'))
+    assert sv**2, sv*sv
+    assert sv**7, sv*sv*sv*sv*sv*sv*sv*sv*sv
```
ericbrts, line 45: I think you mean `assert a == b` and not `assert a, b` which only checks if the tuple `(a, b)` is `True` which it always is.
I checked-out your branch, replaced comma with == and your test isn't passing.

Also, could you check it works for states with annotated photons, and on probability distributions (`BSDistribution`, `SVDistribution`) ?
Thanks!

ericbrts (changes requested):

PR #244, opened 2023-07-24
Pcvl-484 fix pow

MarionQuandela:
Just fixing pow and not using pow from exqalibur

### DIFF
--- a/perceval/utils/statevector.py
+++ b/perceval/utils/statevector.py
@@ -45,7 +45,7 @@ import numpy as np
 import sympy as sp
 
 from exqalibur import FockState, FSArray
-from perceval.utils.qmath import exponentiation_by_squaring
+from .qmath import exponentiation_by_squaring
 
 
 def _fockstate_add(self, other):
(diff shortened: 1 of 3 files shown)


perceval/utils/statevector.py
```diff
@@ -45,6 +45,7 @@
 import sympy as sp
 
 from exqalibur import FockState, FSArray
+from perceval.utils.qmath import exponentiation_by_squaring
```
ericbrts, line 48: When you import from the same package, it's better to use relative imports in order to avoid potential circular dependencies:
`from .qmath import exponentiation_by_squaring`

perceval/utils/statevector.py
```diff
@@ -54,10 +55,12 @@ def _fockstate_sub(self, other):
     return StateVector(self) - other
 
 def _fockstate_pow(self, power: int):
-    bs = self.__copy__()
-    for i in range(power - 1):
-        bs = bs * self
-    return bs
+    if power < 0:
```
ericbrts, line 58: I think you don't need this if, but anyway, this function is to disappear soon because it's been reimplemented C++-side (right?).

MarionQuandela: We need this in case the user does something like bs**-1
But yes, this bit is gonna be replace by the c++ method

PR #246, opened 2023-07-27
Pcvl 456 myqlmconverter

raksharuia:


perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+        :param qlmc: quantum gate-based myqlm circuit
+        :type qlmc: qat.core.Circuit
+        :param use_postselection: when True, uses a `post-processed CNOT` as the last gate. Otherwise, uses only
+            `heralded CNOT`
+        :return: the converted Processor
+        """
+        import qat
+        from qat.core.circuit_builder.matrix_util import circ_to_np
+        # importing the quantum toolbox of myqlm
+        # this nested import fixes automatic class reference generation
```
MarionQuandela, line 68: I do not have any problem if I put the import on the top of the file.
Putting all imports at the top of a python file is more clean.
Moreover if you do so, you can directly type your method declaration:
`def convert(self, qlmc: qat.core.Circuit,`

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+
+from perceval.components import Port, Circuit, Processor, Source, BS, PS
+from perceval.utils import P, BasicState, Encoding
+from perceval.utils.algorithms.optimize import optimize
+from perceval.utils.algorithms.norm import frobenius
+import perceval.components.unitary_components as comp
+import numpy as np
+
+
+min_precision_gate = 1e-4
```
MarionQuandela, line 38: If `min_precision_gate` is a constant, consider putting it in upper case: `MIN_PRECISION_GATE`

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class MyQLMConverter:
+    r"""myQLM quantum circuit to perceval circuit converter.
+    :param catalog: component library of perceval
```
MarionQuandela, line 43: In order to be consistent with the documentation, you must have a blank line between the method definition and the parameters
Moreover the definition of the parameters `backend_name` and `source` are missing

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+            `heralded CNOT`
+        :return: the converted Processor
+        """
+        import qat
+        from qat.core.circuit_builder.matrix_util import circ_to_np
+        # importing the quantum toolbox of myqlm
+        # this nested import fixes automatic class reference generation
+
+        # count the number of CNOT gates to use during the conversion, will give us the number of herald to handle
+        n_cnot = qlmc.count("CNOT")
```
MarionQuandela, line 71: Consider declaring variable only when you need it, and so moving `n_cnot` and `cnot_idx` line 82 after the first for loop

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+            input_list[i * 2] = 1
+        default_input_state = BasicState(input_list)
+
+        for i, instruction in enumerate(qlmc.iterate_simple()):
+            instruction_name = instruction[0]  # name of the Gate
+            instruction_qbit = instruction[-1]  # tuple with list of qbit positions
+            # information carried by instruction
+            # tuple ('Name', [value of the parameter for gate], [list of qbit positions where gate is applied])
+
+            assert instruction_name in qlmc.gate_set, "cannot convert (%s)" % instruction_name
```
MarionQuandela, line 89: Consider using the try/except paradigm instead of assert
Assert verification can be remove at runtime for optimisation (python -O)

MarionQuandela, line 89: You can use f string `f"cannot convert {instruction_name}"` instead of `"cannot convert (%s)" % instruction_name`
I think it's better for code readabily

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+            # only gates are converted -> checking if instruction is in gate_set of AQASM
+
+            if len(instruction_qbit) == 1:
+                if instruction_name == "H":
```
MarionQuandela, line 93: You should declare `ins` (`ins=None`) before the if statement line 93.
First some language work like that (see RAII) and it's a good idea to start coding like that now.
Then if you use this variable elsewhere in your code in the futur you can (and usually will) have unwanted  side effect because your variable will already be assign.
On reading your code, a fellow developper can immediatly spot that this condition statement is about building the right `ins`.

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                    ins = self._create_one_qubit_gate(gate_u)
+                p.add(instruction_qbit[0]*2, ins.copy())
+            else:
+                # only 2 qubit gates
+                c_idx = instruction_qbit[0] * 2
+                c_data = instruction_qbit[1] * 2
+                c_first = min(c_idx, c_data)  # used in SWAP
+
+                if instruction_name == "CNOT":
+                    cnot_idx += 1
```
MarionQuandela, line 111: You should declare `mode_map` (`mode_map = None`) and `cnot_processor` (`cnot_processor = None`) before the if statement line 112
Same reason that `ins` above

EDIT: `mode_map` doesn't need to be instantiate if you consider my two next comments

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                elif instruction_name == "CSIGN":
+                    # Controlled Z in myqlm is named CSIGN
+                    cz_processor = self._heralded_cz_builder.build()
+                    mode_map = {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
+                    p.add(mode_map, cz_processor)
+                elif instruction_name == "SWAP":
+                    # c_idx and c_data are consecutive - not necessarily ordered
+                    p.add(c_first, comp.PERM([2, 3, 0, 1]))
+                else:
+                    raise RuntimeError("Gate not yet supported: %s" % instruction_name)
```
MarionQuandela, line 128: Consider using f strings

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+        cnot_idx = 0
+
+        n_moi = qlmc.nbqbits * 2  # number of modes of interest = 2 * number of qbits
+        input_list = [0] * n_moi
+        p = Processor(self._backend_name, n_moi, self._source)
+
+        for i in range(qlmc.nbqbits):
+            p.add_port(i * 2, Port(Encoding.DUAL_RAIL, f'Q{i}'))
+            input_list[i * 2] = 1
+        default_input_state = BasicState(input_list)
```
MarionQuandela, line 81: `default_input_state` is only used once l129, consider not declaring it and just use `p.with_input(BasicState(input_list))` l129

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+
+        for i, instruction in enumerate(qlmc.iterate_simple()):
+            instruction_name = instruction[0]  # name of the Gate
+            instruction_qbit = instruction[-1]  # tuple with list of qbit positions
+            # information carried by instruction
+            # tuple ('Name', [value of the parameter for gate], [list of qbit positions where gate is applied])
```
MarionQuandela, line 87: Complete line with "qlmc.iterate_simple() is a" tuple ('Name', ...
But it's very good to have put this info here
Consider also putting it a the top of the for loop for readability

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                c_first = min(c_idx, c_data)  # used in SWAP
+
+                if instruction_name == "CNOT":
+                    cnot_idx += 1
+                    if use_postselection and cnot_idx == n_cnot:
+                        cnot_processor = self._postprocessed_cnot_builder.build()
+                        mode_map = {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
+                    else:
+                        cnot_processor = self._heralded_cnot_builder.build()
+                        mode_map = {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
```
MarionQuandela, line 117: `mode_map` should be assign after the if statement, it's the same in both cases

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                # only 2 qubit gates
+                c_idx = instruction_qbit[0] * 2
+                c_data = instruction_qbit[1] * 2
+                c_first = min(c_idx, c_data)  # used in SWAP
+
+                if instruction_name == "CNOT":
+                    cnot_idx += 1
+                    if use_postselection and cnot_idx == n_cnot:
+                        cnot_processor = self._postprocessed_cnot_builder.build()
+                        mode_map = {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
```
MarionQuandela, line 114: `mode_map` should be assign after the if statement, it's the same in both cases

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                    else:
+                        cnot_processor = self._heralded_cnot_builder.build()
+                        mode_map = {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
+                    p.add(mode_map, cnot_processor)
+                elif instruction_name == "CSIGN":
+                    # Controlled Z in myqlm is named CSIGN
+                    cz_processor = self._heralded_cz_builder.build()
+                    mode_map = {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
```
MarionQuandela, line 122: `mode_map` is build 3 times with the exact same pattern (2 if you take in account my comments above), consider using a tiny method to build it:
```
def get_mode_map(c_idx: int, c_data: int) -> dict:
    return {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
```

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                    p.add(c_first, comp.PERM([2, 3, 0, 1]))
+                else:
+                    raise RuntimeError("Gate not yet supported: %s" % instruction_name)
+        p.with_input(default_input_state)
+        return p
+
+    def _create_one_qubit_gate(self, u):
+        # universal method, takes in unitary and approximates one using
+        # Frobenius method
+        if abs(u[1, 0]) + abs(u[0, 1]) < 2 * min_precision_gate:
```
MarionQuandela, line 135: You should declare `ins` (`ins=None`) before the if statement line 135. (see above the comment for the same variable name)

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+                    p.add(mode_map, cz_processor)
+                elif instruction_name == "SWAP":
+                    # c_idx and c_data are consecutive - not necessarily ordered
+                    p.add(c_first, comp.PERM([2, 3, 0, 1]))
+                else:
+                    raise RuntimeError("Gate not yet supported: %s" % instruction_name)
+        p.with_input(default_input_state)
+        return p
+
+    def _create_one_qubit_gate(self, u):
```
MarionQuandela, line 132: Put return type:
 `def _create_one_qubit_gate(self, u) -> Circuit:`

perceval/converters/myqlm_converter.py
```diff
@@ -0,0 +1,151 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from perceval.components import Port, Circuit, Processor, Source, BS, PS
+from perceval.utils import P, BasicState, Encoding
+from perceval.utils.algorithms.optimize import optimize
+from perceval.utils.algorithms.norm import frobenius
+import perceval.components.unitary_components as comp
+import numpy as np
```
MarionQuandela, line 35: Remove unused numpy import (consider using a linter to spot stuff like that)

MarionQuandela (changes requested):
Very nice code to read with plenty of comments (we looooooove comments <3)
I did a lot of comments but it was already pretty good, don't hesitate to spoke if you don't agree with some

tests/test_converter_myqlm.py
```diff
@@ -0,0 +1,243 @@
...
+    import perceval as pcvl
+    pc.with_input(pcvl.LogicalState([0, 0, 0]))
+    sampler = Sampler(pc)
+    output_distribution = sampler.probs()["results"]
+    # GHZ state distribution is expected ( # precision is a bit off because of heralded CNOT implementation)
+    logical000 = BasicState('|1,0,1,0,1,0>')
+    logical111 = BasicState('|0,1,0,1,0,1>')
+    assert output_distribution[logical000] == pytest.approx(0.5, abs=1e-3)
+    assert output_distribution[logical111] == pytest.approx(0.5, abs=1e-3)
+    assert sum(list(output_distribution.values())) == 1
```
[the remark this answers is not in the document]
ericbrts: Actually you don't need to test that our distribution sum to 1... because we're normalizing them anyway, so I guess L.221 assert isn't useful.

perceval/converters/myqlm_converter.py
raksharuia: 👍

perceval/converters/myqlm_converter.py
raksharuia: Implemented with a function outside the class in the script

tests/test_converter_myqlm.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+@pytest.mark.skip(reason="Only for Dev, takes long for computation and displays truth table")
```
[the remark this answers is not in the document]
raksharuia: even though you found it is not too long, it still is implemented to view the results to confirm the test, I skip it and leave it to be tested only if someone knows about it.

MarionQuandela (approved):

PR #250, opened 2023-08-02
PCVL-491 token provider

ericbrts:


tests/test_token_provider.py
```diff
@@ -0,0 +1,58 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from perceval.runtime._token_management import TokenProvider
+
+import os
+
+MISSING_KEY = "MISSING_ENV_VAR"
```
MarionQuandela, line 34: Why define those variable here and not on the test function ?

tests/test_token_provider.py
```diff
@@ -0,0 +1,58 @@
...
+ENV_VAR_KEY = "DUMMY_ENV_VAR"
+TOKEN_FROM_ENV = "DUMMY_TOKEN_FROM_ENV"
+TOKEN_FROM_CACHE = "DUMMY_TOKEN_FROM_CACHE"
+os.environ[ENV_VAR_KEY] = TOKEN_FROM_ENV  # Write a temporary environment variable
+
+def test_token_provider():
+    provider = TokenProvider(MISSING_KEY)
```
MarionQuandela, line 41: Each time a TokenProvider is created in this file, I would do it by naming the parameter `env_var`
`TokenProvider(env_var=MISSING_KEY)`
To emphase that I'm not testing the file feature

perceval/runtime/_token_management.py
```diff
@@ -0,0 +1,88 @@
...
+
+import os
+
+
+class TokenProvider:
+    """Search for a token in different places and retrieve it
+
+    The priority order is:
+    - Token cached in memory
+    - Environment variable
```
MarionQuandela, line 38: add default env variable name in the class description (PCVL_CLOUD_TOKEN)

MarionQuandela (approved):

tests/test_token_provider.py
ericbrts: I want to create constants in this test, and Python convention doesn't like constants from being created inside a function.

PR #251, opened 2023-08-07
Pcvl 490 factor circuit converters code

raksharuia:


perceval/converters/abstract_converter.py
```diff
@@ -0,0 +1,147 @@
...
+
+from abc import ABC, abstractmethod
+
+from perceval.components import Port, Circuit, Processor, Source
+from perceval.utils import P, BasicState, Encoding
+from perceval.utils.algorithms.optimize import optimize
+from perceval.utils.algorithms.norm import frobenius
+import perceval.components.unitary_components as comp
+
+MIN_PRECISION_GATE = 1e-4
```
ericbrts, line 38: Now that we use this value at different places, I suggest we put it with the other global thresholds we use throughout Perceval.
See utils/globals.py

perceval/converters/abstract_converter.py
```diff
@@ -0,0 +1,147 @@
...
+        self._heralded_cz_builder = catalog["heralded cz"]
+        self._postprocessed_cnot_builder = catalog["postprocessed cnot"]
+        self._generic_2mode_builder = catalog["generic 2 mode circuit"]
+        self._lower_phase_component = Circuit(2) // (0, comp.PS(P("phi2")))
+        self._upper_phase_component = Circuit(2) // (1, comp.PS(P("phi1")))
+        self._two_phase_component = Circuit(2) // (0, comp.PS(P("phi1"))) // (1, comp.PS(P("phi2")))
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
```
ericbrts, line 67: I'm not sure what a name brings to the table. It's good to have a name for objects you'd wish to display but we don't want to display a converter.
What do you think?

perceval/converters/abstract_converter.py
```diff
@@ -0,0 +1,147 @@
...
+        """Each converter would have a distinct name as a string"""
+
+    @abstractmethod
+    def num_qbits_gate_circuit(self, gate_circuit) -> int:
```
ericbrts, line 71: the "gate_circuit" words are redundant. I suggest `def count_qubits(self, gate_circuit) -> int:` for readability

perceval/converters/abstract_converter.py
```diff
@@ -0,0 +1,147 @@
...
+        pass
+
+    def configure_processor(self, gate_circuit, **kwargs):
```
ericbrts, line 74: Shouldn't it be a private method? If so, please prefix it with an underscore.

perceval/converters/abstract_converter.py
```diff
@@ -0,0 +1,147 @@
...
+    def apply_input_state(self):
+        default_input_state = BasicState(self._input_list)
+        self._converted_processor.with_input(default_input_state)
+
+    @abstractmethod
+    def convert(self, gate_circuit, use_postselection: bool = True) -> Processor:
+        """
+        converts gates based circuits to one with linear optical components
+        and returns a perceval processor
+        Children - MyQlMConverter and QiskitConverter should define this method
```
ericbrts, line 96: Please don't name children in a parent class. Being abstract means you need to be generic, and have new children without knowing about it.
Also, the info "Children should define this method" is already clear with the `@abstractmethod` modifier

perceval/converters/myqlm_converter.py
```diff
@@ -108,51 +87,19 @@ def convert(self, qlmc, use_postselection: bool = True) -> Processor:
                     gate_id = qlmc.ops[i].gate
                     gate_matrix = qlmc.gateDic[gate_id].matrix  # gate matrix data
                     gate_u = circ_to_np(gate_matrix)  # gate matrix to numpy
-                    ins = self._create_one_qubit_gate(gate_u)
-                p.add(instruction_qbit[0]*2, ins.copy())
+                    ins = super()._create_generic_1_qubit_gate(gate_u)
```
ericbrts, line 90: Calling `super().__init__` in children `__init__` methods is basically what you want (i.e. call the parent initializer from the child). However, for other calls, it's not what you want to do. What you want is:
`ins = self._create_generic_1_qubit_gate(gate_u)`
which means use the `_create_generic_1_qubit_gate` method from the parent, *unless the child has overridden it*, which is good practice (that's related to the L from the SOLID principles : https://en.wikipedia.org/wiki/Liskov_substitution_principle - you have to be able to change the behavior of a system by implementing a different child class).

perceval/converters/myqlm_converter.py
```diff
@@ -108,51 +87,19 @@ def convert(self, qlmc, use_postselection: bool = True) -> Processor:
...
-                if abs(u[1, 1] - 1) < MIN_PRECISION_GATE:
-                    ins = self._lower_phase_component.copy()
-                else:
-                    ins = self._two_phase_component.copy()
-            optimize(ins, u, frobenius, sign=-1)
-        else:
-            ins = self._generic_2mode_builder.build_circuit()
-            optimize(ins, u, frobenius, sign=-1)
-        return ins
+                super()._create_2_qubit_gates_from_catalog(instruction_name, n_cnot, c_idx, c_data,
```
ericbrts, line 101: super() -> self
See my other comment

perceval/converters/qiskit_converter.py
```diff
@@ -91,51 +76,18 @@ def convert(self, qc, use_postselection: bool = True) -> Processor:
 
             if instruction[0].num_qubits == 1:
                 # one mode gate
-                ins = self._create_one_qubit_gate(instruction[0].to_matrix())
+                ins = super()._create_generic_1_qubit_gate(instruction[0].to_matrix())
```
ericbrts, line 79: super() -> self
See my other comment

perceval/converters/qiskit_converter.py
```diff
@@ -91,51 +76,18 @@ def convert(self, qc, use_postselection: bool = True) -> Processor:
...
-                if abs(u[1, 1] - 1) < min_precision_gate:
-                    ins = self._lower_phase_component.copy()
-                else:
-                    ins = self._two_phase_component.copy()
-            optimize(ins, u, frobenius, sign=-1)
-        else:
-            ins = self._generic_2mode_builder.build_circuit()
-            optimize(ins, u, frobenius, sign=-1)
-        return ins
+                super()._create_2_qubit_gates_from_catalog(instruction[0].name, n_cnot, c_idx, c_data, c_first,
```
ericbrts, line 90: super() -> self
See my other comment

perceval/converters/abstract_converter.py
```diff
@@ -0,0 +1,147 @@
...
+    return {c_idx: 0, c_idx + 1: 1, c_data: 2, c_data + 1: 3}
+
+
+class AGateConverter(ABC):
+    r"""
+    Converter class for gate based Circuits to perceval processor
+    Qiskit or MyQLM
+    """
+
+    def __init__(self, catalog, **kwargs):
```
ericbrts, line 51: Since `MyQLMConverter` and `QiskitConverter` `__init__` methods share exactly the same signature (i.e. same parameters, same order, same names), I don't know why you went for **kwargs.
**kwargs is a solution for signature variability, but here, signatures are exactly the same.

ericbrts (changes requested):

perceval/converters/abstract_converter.py
raksharuia: I should confess I wasn't sure of its use myself but I used a similar structure from class `ABackend`.  If it actually is never used at all, I can remove it. Why do we have them in backends?

On a similar note, why do we have `preferred command()` in backend classes?

perceval/converters/myqlm_converter.py
raksharuia: Thanks a lot for this explanation :-). Fixed.

raksharuia (commented):
Thanks for all the explanations on code structure - very helpful in understanding the purposes and practices of abstraction and OOP. I only have one question concerning the method `name`. Fixed all the other changes :-)

ericbrts (approved):

PR #252, opened 2023-08-09
PCVL-471 add toffoli gate

MarionQuandela:


perceval/components/core_catalog/toffoli.py
```diff
@@ -0,0 +1,85 @@
...
+ctrl1 (dual rail) ─────┤     ├───── ctrl1 (dual rail)
+                  ─────┤     ├─────
+                       │     │
+data (dual rail)  ─────┤     ├───── data (dual rail)
+                  ─────┤     ├─────
+                       ╰─────╯"""
+
+    def __init__(self):
+        super().__init__("toffoli")
+        self._default_opts['type'] = AsType.PROCESSOR
```
ericbrts, line 50: You don't need this line as it's part of the deprecated builder pattern.

perceval/components/core_catalog/toffoli.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @deprecated(version="0.10.0", reason="Use build_circuit or build_processor instead")
```
ericbrts, line 52: Don't bother implementing deprecated methods in new classes. I removed the abstractmethod modifier so that we can get rid of it in new components (see mzi.py for instance)

PR #254, opened 2023-08-09
PCVL-493 apply values to interferometer phases

ericbrts:


raksharuia (commented):
Cool rework on the interferometer. 

Some of the features caused issues on my local machine (Summary below).

1. DebugSkin did not work.
2. I could not use Debugger in tests to view and analyze variable information of interferometer object.

perceval/rendering/circuit/debug_skin.py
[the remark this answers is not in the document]
ericbrts: From what I see, you should've passed an instanciated DebugSkin object, not just the type :
```python
pcvl.pdisplay(small_interferometer, skin=DebugSkin())  # With parenthesis :)
```

tests/test_generic_interferometer.py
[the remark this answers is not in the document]
ericbrts: I'm really not sure how I can help you on that one.

tests/test_vizualization.py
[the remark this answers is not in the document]
ericbrts: You're right, we didn't pick a convention on how to import, when Perceval was started.

I personally go for:
```python
from perceval.sub_package import A, B, C, D
```
most of the time. We can talk about this.

But I'm improving only the lines/files I touch, and not all at once (in order to avoid big unreadable diffs in the history).

perceval/components/generic_interferometer.py
```diff
@@ -0,0 +1,193 @@
...
+    def set_params_from_other(self, other: Circuit, top_left_pos: Tuple[int, int]):
+        """Retrieve parameter value from another interferometer
+
+        :param other: Another interferometer
+        :param top_left_pos: Starting position of the insertion. See full description in `set_param_list`
+        """
+        if not other.defined:
+            raise ValueError("Cannot copy parameters from a circuit which isn't fully defined")
+        param_list = [float(p) for p in other.get_parameters()]
+        self.set_param_list(param_list, top_left_pos, other.m)
```
MarionQuandela, line 193: Consider adding a __repr__ method to display some infos in debug mode (since the interferometer cannot be displayed)

    def __repr__(self):
        return f"shape:{self._shape}, depth:{self._depth}, depth_per_mode:{self._depth_per_mode}, has_input_phase_layer:{self._has_input_phase_layer}"

perceval/components/linear_circuit.py
```diff
@@ -581,11 +549,11 @@ def decomposition(U: MatrixN,
...
-            if shape == "triangle":
+            if shape == InterferometerShape.TRIANGLE:
                 lc = decomposition.decompose_triangle(U, component, phase_shifter_fn, permutation, precision,
                                                       constraints, allow_error=allow_error,
                                                       ignore_identity_block=ignore_identity_block)
-            else:
+            elif shape == InterferometerShape.RECTANGLE:
                 lc = decomposition.decompose_rectangle(U, component, phase_shifter_fn, permutation, precision,
                                                        constraints, allow_error=allow_error,
                                                        ignore_identity_block=ignore_identity_block)
```
MarionQuandela, line 559: consider handle the else case (with a not implemented exception for exemple) if you plan one day of enriched the InterferometerShape enum

perceval/components/linear_circuit.py
```diff
@@ -537,7 +500,7 @@ def copy(self, subs: Union[dict,list] = None):
     def decomposition(U: MatrixN,
```
MarionQuandela, line 500: Don't we want this method to explicitly fail instead of returning a "None" ?
Consider adding a warning at the end of the method in this case "cannot decompose this circuit ..."

MarionQuandela (approved):

perceval/components/linear_circuit.py
ericbrts: Done

perceval/components/generic_interferometer.py
ericbrts: Done, with minimal info

perceval/components/generic_interferometer.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class GenericInterferometer(Circuit):
+    r"""Generate a generic interferometer circuit with generic elements and optional phase_shifter layer
+
+    :param m: number of modes
+    :param fun_gen: generator function for the building components, index is an integer allowing to generate
+                    named parameters - for instance:
+                    :code:`fun_gen=lambda idx: phys.BS()//(0, phys.PS(pcvl.P(f"phi_{idx}")))`
+    :param shape: The output interferometer shape (triangle or rectangle)
```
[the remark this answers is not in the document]
ericbrts: Done

PR #255, opened 2023-08-10
Pcvl 502 rework logical state

MarionQuandela:


perceval/components/port.py
```diff
@@ -158,31 +190,11 @@ def is_connected_to(self, component) -> bool:
...
-                    index += 1
-                elif port.encoding == Encoding.QUDIT or port.encoding == Encoding.POLARIZATION:
-                    raise NotImplementedError
-        if index != len(self):
-            raise ValueError('Logical state and port list do not match (state too long)')
-        return BasicState(result)
-
-    def __str__(self):
-        return '|' + ','.join([str(x) for x in self]) + '>_L'
+def get_BS_from_ports(ports : list[APort], state: LogicalState) -> BasicState:
```
ericbrts, line 193: BS is misleading as it also refers to beam splitters. What would you think of calling this function `get_state_from_ports` ?

perceval/components/port.py
```diff
@@ -81,6 +87,24 @@ def encoding(self):
     def is_output_photonic_mode_closed(self):
         return False
 
+    @staticmethod
+    def has_logical_state_equivalent() -> bool:
+        return True
+
+    def to_logic_state(self, state : bool) -> LogicalState:
```
ericbrts, line 94: I think this method does somewhat the opposite of what it says: why does `state` = 0 outputs |0,1>_Logical? Doesn't make sense to me.

Also, I don't like the signature: why is `state` a boolean? It's not explained in the docstring.

Also, the term _logical_ is used everywhere else. Why _logic_ here?

perceval/components/port.py
```diff
@@ -158,31 +190,11 @@ def is_connected_to(self, component) -> bool:
...
-    def __str__(self):
-        return '|' + ','.join([str(x) for x in self]) + '>_L'
+def get_BS_from_ports(ports : list[APort], state: LogicalState) -> BasicState:
+    logical_state = LogicalState()
+    port_list = [port for port in ports if port.has_logical_state_equivalent()]
+    if len(port_list) != len(state):
+        raise ValueError('Logical state and port list size do not match')
+    for port, mode in zip(port_list, state):
+        logical_state += port.to_logic_state(mode)
+    return BasicState(logical_state)
```
ericbrts, line 200: I think there's a misunderstanding. This line should not happen, because it's semantically wrong. Why "casting" a logical state to a basic state: they represent different things and are not equal for all encodings except one (RAW).

ericbrts (changes requested):

tests/test_port.py
```diff
@@ -60,3 +61,10 @@ def test_digital_converter():
     detector.trigger(False)
     assert not phi.is_symbolic()
     assert float(phi) == np.pi/4
+
+
+def test_logical_conversion():
+    ports = [Herald(1), Port(Encoding.DUAL_RAIL, "belle"), Port(Encoding.RAW, "bulle"), Herald(1), Herald(0), Port(Encoding.TIME, "rebelle"), Herald(1)]
```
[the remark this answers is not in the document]
MarionQuandela: No because the Herald are not a "true" state since it cannot be changed (I rename some methods and I think it's clearer now)

perceval/components/port.py
```diff
@@ -91,13 +92,15 @@ def is_output_photonic_mode_closed(self):
     def has_logical_state_equivalent() -> bool:
         return True
 
-    def to_logic_state(self, state : bool) -> LogicalState:
-        """Return the logical state corresponding to a state
+    def to_state_with_encoding(self, state : int) -> LogicalState:
```
ericbrts, line 95: `def to_basic_state(self, qubit_state: int) -> BasicState`

perceval/components/port.py
```diff
@@ -190,11 +193,19 @@ def is_connected_to(self, component) -> bool:
         return component in self._connections
 
 
-def get_BS_from_ports(ports : list[APort], state: LogicalState) -> BasicState:
+def get_state_with_encoding_from_ports(ports : List[APort], state: LogicalState) -> LogicalState:
```
ericbrts, line 196: get_basic_state_from_ports() -> BasicState

+ ajouter bool (default False) pour prendre en compte les ancillaries

ericbrts (approved):

PR #257, opened 2023-08-21
Pcvl 469 gates unit test

MarionQuandela:


ericbrts (approved):

perceval/components/abstract_processor.py
```diff
@@ -448,6 +448,28 @@ def add_port(self, m, port: APort, location: PortLocation = PortLocation.IN_OUT)
             self._out_ports[port] = port_range
         return self
 
+    def remove_port(self, m, location: PortLocation = PortLocation.IN_OUT):
+        if location in (PortLocation.IN_OUT, PortLocation.INPUT):
```
ericbrts, line 452: Could you factor **in** vs **out** port handling ? Both `if` are basically copy/paste of each other.

perceval/components/core_catalog/toffoli.py
```diff
@@ -26,42 +26,34 @@
...
-                       │     │
-data (dual rail)  ─────┤     ├───── data (dual rail)
-                  ─────┤     ├─────
-                       ╰─────╯"""
+    description = r"""Toffoli gate CCNOT gate with 6 heralded modes and a post-selection function (built using Heralded CCZ and H)."""
+    str_repr = r"""                               ╭──────────╮
+ctrl0 (dual rail) ─────────────┤          ├───────────── ctrl0 (dual rail)
+                  ─────────────┤          ├─────────────
+ctrl1 (dual rail) ─────────────┤          ├───────────── ctrl1 (dual rail)
+                  ─────────────┤ Heralded ├─────────────
```
ericbrts, line 41: Toffoli is not based on a "Heralded CZ" but on a "Post-processed CCZ"!

tests/test_2qbits_gates.py
```diff
@@ -0,0 +1,162 @@
...
+    analyzer_klm_cnot = Analyzer(catalog["klm cnot"].build_processor(), state_dict)
+    analyzer_klm_cnot.compute(expected=expected)
+
+    assert pytest.approx(analyzer_klm_cnot.fidelity, 1E-4) == 1
+
+    # Postprocessed CNOT
+    analyzer_postprocessed_cnot = Analyzer(catalog["postprocessed cnot"].build_processor(), state_dict)
+    analyzer_postprocessed_cnot.compute(expected=expected)
+
+    assert analyzer_postprocessed_cnot.fidelity == 1
```
ericbrts, line 56: Don't you miss a `pytest.approx` here?

tests/test_2qbits_gates.py
```diff
@@ -0,0 +1,162 @@
...
+                else:
+                    assert pytest.approx(rpa, error) == -value
+                    assert rpa < 0
+                assert pytest.approx(ipa) == 0
+            else:
+                assert pytest.approx(rpa) == 0
+                assert pytest.approx(ipa) == 0
+
+
+def test_cz_phase():
```
ericbrts, line 102: Could you document how phases are tested? I feel like they're also tested in the previous test. What's new in this one?

ericbrts (approved):

perceval/components/core_catalog/toffoli.py
MarionQuandela: Actually It's a heralded & post-processed ccz but i will change to just "ccz"

tests/test_2qbits_gates.py
MarionQuandela: No because our postprocessed_cnot is exacly one, we don't need the approx

raksharuia: I just noticed that we are using the same method for all our different CNOT gates, maybe we can simplify by using `pytest.parametrize`?

tests/test_2qbits_gates.py
```diff
@@ -0,0 +1,162 @@
...
+                    assert pytest.approx(rpa, error) == value
+                    assert rpa > 0
+                else:
+                    assert pytest.approx(rpa, error) == -value
+                    assert rpa < 0
+                assert pytest.approx(ipa) == 0
+            else:
+                assert pytest.approx(rpa) == 0
+                assert pytest.approx(ipa) == 0
+
```
GdGliniasty, line 100: I think that on this whole fonction you are ensuring a special case where the amplitudes are real. Thus you exclude benign global phases that you can have in some optical implementation.  So this works because your amplitudes will be exactly +1/3 or -1/3. 
Be careful: if you have some amplitudes that are +/- 1/3 + ipa and ipa=0, then your gate can pass the test but not be a CZ. 
If you want to be more general ( but still ensuring the CZ works), you should look at the exponential form : r.exp(i.θ) . 
If you have this you have to ensure that:
_ if input==output then r=value else r=0 
_if input==ouput and input != data_state then θ=phase_value 
_if input==ouput and input == data_state then θ=phase_value +pi ( mod 2pi ?)

tests/test_2qbits_gates.py
```diff
@@ -0,0 +1,162 @@
...
+    processor.add(2, BS.H())
+    processor.add(0, catalog["postprocessed cnot"].build_processor())
+    processor.add(2, BS.H())
+    check_cz_with_heralds(processor, BasicState("|0,0>"))
+
+
+@pytest.mark.skip(reason="redundant with overhead test")
+@pytest.mark.parametrize("cnot_gate", ["klm cnot", "postprocessed cnot", "heralded cnot"])
+def test_inverse_cnot(cnot_gate):
+    """Test cnot gate phase by inverting it with Hadamard gates
```
GdGliniasty, line 127: This doesn't cost much but is not worth much in comparaison of testing probability amplitudes of the CNOT gate.

tests/test_3qbits_gates.py
```diff
@@ -0,0 +1,145 @@
...
+from perceval.algorithm import Analyzer
+from perceval.utils import generate_all_logical_states
+from perceval.simulators import SimulatorFactory
+from perceval.components import catalog, BS, get_basic_state_from_ports
+
+
+def check_czz_with_heralds(processor, herald_states, error=1E-6):
+    """Check if the CZZ is correct
+
+    :param processor: CZZ processor to check, we assume that ctrl is on [0,1]
```
GdGliniasty, line 42: Same comment than with the check_cz_with_heralds

tests/test_2qbits_gates.py
raksharuia: is the fidelity of the heralded_cnot also exactly = 1 ? or maybe we need pytest.approx here?

tests/test_2qbits_gates.py
```diff
@@ -0,0 +1,162 @@
...
+    processor = Processor("SLOS", 4)
+    processor.add(2, BS.H())
+    processor.add(0, catalog["postprocessed cnot"].build_processor())
+    processor.add(2, BS.H())
+    check_cz_with_heralds(processor, BasicState("|0,0>"))
+
+
+@pytest.mark.skip(reason="redundant with overhead test")
+@pytest.mark.parametrize("cnot_gate", ["klm cnot", "postprocessed cnot", "heralded cnot"])
+def test_inverse_cnot(cnot_gate):
```
raksharuia, line 126: Comment in this test states that it tests for phases for inverted CNOT gate but the method only checks fidelity and does not have any check for phase. Maybe change the comment?

tests/test_3qbits_gates.py
```diff
@@ -0,0 +1,145 @@
...
+import numpy as np
+
+from perceval import BasicState, Port, Processor, Encoding, PostSelect
+from perceval.algorithm import Analyzer
+from perceval.utils import generate_all_logical_states
+from perceval.simulators import SimulatorFactory
+from perceval.components import catalog, BS, get_basic_state_from_ports
+
+
+def check_czz_with_heralds(processor, herald_states, error=1E-6):
```
raksharuia, line 39: czz or ccz?

raksharuia (commented):
Cool work

tests/test_2qbits_gates.py
raksharuia: @GdGliniasty, how would a global phase affect? Would it not be factored out?

GdGliniasty: Indeed it would be factored out, but this test doesn't allow it. Let say your gate is diag(i,i,i,-i). This is basically iCZ . For in_state=out_state=|00>, rpa=0 and i=1 then value =0 and and the lign "assert" rpa>0" breaks the test.

PR #259, opened 2023-08-23
Pcvl 496 better heralded cnot

raksharuia:


tests/test_cnot.py
ericbrts, line 1: As discussed a few days ago, this test should be merged with the gate unit test @MarionQuandela is writing.
Could you both handle it, please?

perceval/components/core_catalog/heralded_cnot.py
```diff
@@ -26,28 +26,25 @@
...
-                 ─────┤     ├─────
-                      ╰─────╯"""
-    see_also = "postprocessed cnot"
-
-    R1 = 0.228
-    R2 = 0.758
-    theta1 = BS.r_to_theta(R1)
-    theta2 = BS.r_to_theta(R2)
+    article_ref = ""
+    description = r"""Heralded CNOT gate built using Heralded CZ and H"""
```
ericbrts, line 39: Could you please state that there are only 2 heralded modes in this version of the CNOT gate?
Like:
```python
description = r"""CNOT gate with 2 heralded modes"""
```

perceval/components/core_catalog/heralded_cnot.py
```diff
@@ -61,29 +58,19 @@ def build(self):
...
 
     def build_processor(self, **kwargs):
-        p = self._init_processor(**kwargs)
-        return p.add_port(0, Port(Encoding.DUAL_RAIL, 'ctrl')) \
-            .add_port(2, Port(Encoding.DUAL_RAIL, 'data')) \
-            .add_herald(4, 0) \
-            .add_herald(5, 1) \
-            .add_herald(6, 0) \
-            .add_herald(7, 1)
+        p = Processor(backend="SLOS", m_circuit=4, source=Source())
```
ericbrts, line 69: You are bypassing the `self._init_processor(**kwargs)` method call.
This means :
* The circuit building code is duplicated
* You hardcoded that you're using SLOS (_init_processor creates a Processor with any backend when kwargs require it)

So, could you get back to how other catalog items create their processors?

tests/test_cnot.py
MarionQuandela: You can merge it and I will adapt my branch with your tests

PR #266, opened 2023-09-04
Pcvl 489 fix mps backend

raksharuia:


perceval/backends/_mps.py
```diff
@@ -41,26 +40,36 @@
...
 
 class MPSBackend(AProbAmpliBackend):
-    """Step-by-step circuit propagation algorithm, works on a circuit.
-    Approximate the probability amplitudes with a cutoff.
+    """
+    The state of the system is written in form of an MPS and
+    updated step-by-step by a circuit propagation algorithm.
+
+    Approximate the probability amplitudes with a cutoff -> bond Dimension in an MPS.
     - For now only supports Phase shifters and Beam Splitters
```
ericbrts, line 48: Shouldn't it be reworded to "supports components up to two modes", after what we've seen in the code?

perceval/backends/_mps.py
```diff
@@ -70,149 +79,230 @@ def set_circuit(self, circuit: ACircuit):
         for r, c in C:
             assert c.compute_unitary(use_symbolic=False).shape[0] <= 2, \
                 "MPS backend can not be used with components of using more than 2 modes"
-        if self._cutoff is None:
-            self._cutoff = C.m
 
     def set_input_state(self, input_state: BasicState):
         super().set_input_state(input_state)
-        self.compile()
+        self._cutoff = input_state.n + 1  # sets the default value of cut-off to max number of photons
```
ericbrts, line 85: Is `input_state.n + 1` the minimum cut off value that makes MPS work?
If so, you shouldn't impose this value here, as it is overriding any value that could be set by the user, through the `__init__` or `set_cutoff` methods.

Maybe you could handle this cut off minimum value in `compile` where the maximum is already dealt with (see my other comment)

perceval/backends/_mps.py
```diff
@@ -70,149 +79,230 @@ def set_circuit(self, circuit: ACircuit):
...
-        # self._input_state *= BasicState([0] * (self._input_state.m - self._input_state.m))
-        self.n = self._input_state.n
-        self.d = self.n + 1
-        self._cutoff = min(self._cutoff, self.d ** (self._input_state.m//2))
-        self._gamma = np.zeros((self._input_state.m, self._cutoff, self._cutoff, self.d), dtype='complex_')
+        self._current_input = None  # todo: ERIC - do I need to set it to None again?
+
+        self._n = self._input_state.n  # total number of photons
+        self._d = self._n + 1  # possible num of photons in each mode {0,1,2,...,n}
+        self._cutoff = min(self._cutoff, self._d ** (self._input_state.m//2))
```
ericbrts, line 122: Here, max cutoff is dealt with. Maybe deal with min value too on this line (?)

perceval/backends/_mps.py
raksharuia: There were some issues with setting up and clearing of previous cut_off values. Some reset (future dev) method for attributes of backends maybe good. I found some solution to not override and also perform checks on cutoff.

ericbrts (approved):

PR #268, opened 2023-09-07
PCVL-524 fix non-superposed state vector normalization

ericbrts:


tests/test_polarization_simulator.py
```diff
@@ -118,7 +118,7 @@ def test_polarization_circuit_0():
         res = psimu.probs(BasicState("|{P:H}>"))
```
MarionQuandela, line 118: You can parametrize this test
```
@pytest.mark.parametrize("backend_name", ["Naive", "SLOS", "MPS"])
def test_polarization_circuit_0(backend_name):
...
```

MarionQuandela (approved):

PR #287, opened 2023-10-11
Pcvl 541 adapt ci to gedik notebook

MarionQuandela:


docs/source/notebooks/Gedik_qudit.ipynb
```diff
@@ -571,40 +559,28 @@
...
-    "data_qpu.append(res_0.get_results()['results'])\n",
-    "data_qpu.append(res_1.get_results()['results'])\n",
-    "data_qpu.append(res_2.get_results()['results'])\n",
-    "data_qpu.append(res_3.get_results()['results'])\n",
-    "data_qpu.append(res_4.get_results()['results'])\n",
-    "data_qpu.append(res_5.get_results()['results'])\n",
-    "data_qpu.append(res_6.get_results()['results'])\n",
-    "data_qpu.append(res_7.get_results()['results'])"
+    "for i in range(8):\n",
+    "    data_qpu.append(jobs[i].get_results()['results'])"
```
ericbrts, line 578: Jobs are ran asynchronously so they might run for long and the notebook would go on. Is there anything checking that the jobs have finished before the `get_results()` call ? If not, I guess the notebook is not re-runnable by a user, as it would require an arbitrary waiting time between this cell and the previous one.

ericbrts (approved):

PR #298, opened 2023-10-20
0 photons evolution supported

melvinmathe:


perceval/simulators/simulator.py
```diff
@@ -388,6 +388,9 @@ def evolve(self, input_state: Union[BasicState, StateVector]) -> StateVector:
         for probampli, instate_list in decomposed_input:
             reslist = []
             for in_s in instate_list:
+                if in_s.n == 0:
```
MarionQuandela, line 391: Can you also add a simple UT for this case in tests/test_simulator.py please ?

melvinmathe: Yes of course

PR #300, opened 2023-10-23
Pcvl 573 analyzer

MarionQuandela:


docs/source/tools.rst
```diff
@@ -172,7 +172,7 @@ Then, we display the result of `Circuit Analyzer` via ``pdisplay``.
 
 >>> pcvl.pdisplay(ca)
 
-.. figure:: _static/img/CircuitAnalyserHOM.png
```
ericbrts, line 175: I think you should rollback this change, else the image link is broken

MarionQuandela: No because I also renamed the png file

ericbrts: Right, sorry

PR #302, opened 2023-10-26
Pcvl 571 rerun and check mps notebooks

raksharuia:
Removed all todo's and made a note of possible future tasks in this jira ticket


perceval/backends/_mps.py
```diff
@@ -144,8 +143,6 @@ def _compile(self) -> bool:
         # This initialization of MPS (gamma and sv) fixes the input state to be completely separable
         # and a pure BasicState (no superposition); hence would have only 1 non-zero element whose value = 1.
         # It is simply written based on this choice as the SVD of such a structure would exactly look like this
-        # todo: initialization of MPS could be more generic - to include mixed/superposed states as input
```
MarionQuandela, line 147: Did we drop this idea ? If not is there a JIRA ticket that exists ?

perceval/backends/_mps.py
```diff
@@ -284,7 +280,6 @@ def _transition_matrix_2_mode(self, u):
         u11, u12, u21, u22 = u[0, 0], u[0, 1], u[1, 0], u[1, 1]
         d = self._d
         big_u = np.zeros((d, d, d, d), dtype='complex_')  # matrix corresponding to action of BS on the 2 modes
-        # todo: find a way to vectorize and remove "for" loops
```
MarionQuandela, line 287: Are we dropping this ?

MarionQuandela (commented):
Are the todo that you removed are done (or won't fix ?) ?

perceval/backends/_mps.py
raksharuia: I updated this as a comment on the current JIRA ticket. We need to decide if we want to create tickets now or keep this information elsewhere

perceval/backends/_mps.py
raksharuia: again a refinement - should be a task combined with putting this backend section into Quandelibc like we have for SLOS, for example

raksharuia:
The todo's that are to be done are as comments in the JIRA ticket PCVL-571. One or two were questions to Rawad - I discussed with him and we do not nee to change anything now

PR #303, opened 2023-10-27
PCVL-480-handle-partial-results-job-status

MarionQuandela:


perceval/runtime/job_status.py
```diff
@@ -143,6 +143,10 @@ def success(self):
     def failed(self):
         return self._status in [RunningStatus.CANCELED, RunningStatus.ERROR]
 
+    @property
+    def maybe_completed(self):
+        return self._status in [RunningStatus.PARTIAL_COMPLETION, RunningStatus.UNKNOWN]
```
ericbrts, line 148: The new PARTIAL_COMPLETION status is a valid case of completion. The "partial" part stands for "you don't have all the data you asked for", but the job is complete and in success.
So PARTIAL_COMPLETION  shouldn't be in maybe_completed, but in completed

perceval/runtime/local_job.py
```diff
@@ -116,17 +117,29 @@ def execute_async(self, *args, **kwargs) -> Job:
...
+                raise KeyError("Cannot find either 'result' or 'results_list' in self._results")
+
+    def get_results(self) -> Dict:
+        if not self.is_complete and not self.is_maybe_complete:
+            raise RuntimeError('The job is still running, results are not available yet.')
+        job_status = self.status
+        if job_status.status == RunningStatus.SUCCESS:
+            self._assign_results()
+        elif self.is_maybe_complete:
+            warnings.warn("Partial or unknown job status, trying to get result anyway.")
```
ericbrts, line 138: Remote "Partial or" (?)

perceval/runtime/local_job.py
```diff
@@ -116,17 +117,29 @@ def execute_async(self, *args, **kwargs) -> Job:
...
+            try:
+                self._assign_results()
+            except KeyError:
+                return {}
```
ericbrts, line 142: Shouldn't we raise, here? The except would allow us to create a brand new user-readable error message (?)

perceval/runtime/remote_job.py
```diff
@@ -199,4 +194,25 @@ def get_results(self) -> Any:
                     res["results"] = result_mapping_function(res["results"], **self._delta_parameters)
             else:
                 results["results"] = result_mapping_function(results["results"], **self._delta_parameters)
+
+    def get_results(self) -> Dict:
+        job_status = self.status
+        if not job_status.completed and not job_status.maybe_completed:
+            raise RuntimeError('The job is still running, results are not available yet.')
+
+        if not self.is_maybe_complete and job_status.status != RunningStatus.SUCCESS:
```
ericbrts, line 203: Use `job_status.maybe_completed` for consistency?

perceval/runtime/remote_job.py
```diff
@@ -199,4 +194,25 @@ def get_results(self) -> Any:
...
+        if job_status.status == RunningStatus.SUCCESS:
+            response = self._rpc_handler.get_job_results(self._id)
+            results = deserialize(json.loads(response)['results'])
+        elif self.is_maybe_complete:
+            warnings.warn("Partial or unknown job status, trying to get result anyway.")
+            try:
+                response = self._rpc_handler.get_job_results(self._id)
+                results = deserialize(json.loads(response)['results'])
+            except KeyError:
+                return {}
```
ericbrts, line 216: Same as for the local job: if we return an empty dictionary, we should as well raise an exception.

ericbrts (changes requested):

PR #304, opened 2023-10-30
PCVL-553 pdisplay for tomography

MarionQuandela:


perceval/rendering/pdisplay.py
```diff
@@ -219,12 +226,89 @@ def pdisplay_state_distrib(sv: Union[StateVector, ProbabilityDistribution, BSCou
...
+
+    basis = []
+    for val in pauli_names:
+        basis.append(''.join(val))
+    return basis
+
+
+def _get_sub_figure(ax, array, basis_name):
+    # Data
+    size = array.itemsize * 2
```
raksharuia, line 247: `array.itemsize` is not the size or shape of the 2D matrix called "array". 
(For definition See fig or doc for more details: [https://docs.python.org/fr/3.10/library/array.html])

Screenshot 2023-10-31 at 11 54 29


My guess is that you want the size of the 2D matrix :

For example, a 2-qubit gate, like CNOT, chi would always be a 16x16 matrix). To retrieve this, you need (works on numpy arrays, even numpy.ndarray)-

`array.shape`

output exapmles: (16x16) for 2 qubit gate, (4x4) for 1 qubit gate, (64x64) for 3 qubit gate

Doing `array.itemsize` will give you 8 always for all the three qubit cases above. I think the error did not appear as you were in 2 qubit case.

perceval/rendering/pdisplay.py
```diff
@@ -219,12 +226,89 @@ def pdisplay_state_distrib(sv: Union[StateVector, ProbabilityDistribution, BSCou
...
+    x = numpy.array([[i] * size for i in range(size)]).ravel()  # x coordinates of each bar
+    y = numpy.array([i for i in range(size)] * size)  # y coordinates of each bar
+    z = numpy.zeros(size * size)  # z coordinates of each bar
+    dxy = numpy.ones(size * size) * 0.5  # Width/Lenght of each bar
+    dz = array.ravel()  # length along z-axis of each bar (height)
+
+    # Colors
+    color_map = plt.cm.get_cmap('viridis_r')
```
raksharuia, line 255: not going the colormap `cool`?

perceval/rendering/pdisplay.py
```diff
@@ -258,6 +342,22 @@ def _pdisplay(bsc, **kwargs):
...
+    return simple_float(f, **opts_simple)[1]
+
+
+@dispatch(complex)
+def _pdisplay(c, **kwargs):
+    opts_simple = {}
+    if "precision" in kwargs:
+        opts_simple["precision"] = kwargs["precision"]
+    return simple_complex(c, **opts_simple)[1]
+
```
raksharuia, line 359: I am sorry, I forgot why were these `@dispatch(complex)` and `@dispatch(float)` implemented. Please explain it to me.

I see it replaces the implementation inside `_pdisplay()`, but why is this better? Do we want to use `_pdisplay` for simply float and complex numbers?

raksharuia (changes requested):

perceval/rendering/pdisplay.py
MarionQuandela: No it's not color blind friendly :'(

perceval/rendering/pdisplay.py
MarionQuandela: `dispatch` decorator allow to change the method executed depending of the type of the parameter (you can see that there is 10 method_pdisplay in the same file, this would not be possible without the decorator dispatch) so there is actually a _pdisplay for each supported type.
You can read more about it here:
https://www.geeksforgeeks.org/dispatch-decorator-in-python/

perceval/rendering/pdisplay.py
```diff
@@ -258,6 +342,22 @@ def _pdisplay(bsc, **kwargs):
     return pdisplay_state_distrib(bsc, **kwargs)
 
 
+@dispatch(float)
+def _pdisplay(f, **kwargs):
+    opts_simple = {}
+    if "precision" in kwargs:
+        opts_simple["precision"] = kwargs["precision"]
+    return simple_float(f, **opts_simple)[1]
```
ericbrts, line 350: I'm pretty sure it would be better to pass `**kwargs` to `simple_float` than build a `opts_simple` dictionary. Same in the complex signature.

MarionQuandela: implement a fix

perceval/rendering/pdisplay.py
MarionQuandela: Fixed it thanks !

perceval/rendering/pdisplay.py
raksharuia: Too bad. Then let's keep the color-blind friendly one as default as you had.

perceval/rendering/pdisplay.py
```diff
@@ -219,12 +226,89 @@ def pdisplay_state_distrib(sv: Union[StateVector, ProbabilityDistribution, BSCou
...
+
+def pdisplay_tomography_chi(qpt, output_format: Format = Format.MPLOT, precision=1E-6):
+    if output_format != Format.MPLOT:
+        raise TypeError("Tomography plot only support MPLOT")
+
+    chi_op = qpt.chi_matrix()
+
+    fig = plt.figure()
+    pauli_captions = _generate_pauli_captions(qpt._nqubit)
+    significant_digit = int(math.log10(1 / precision))
```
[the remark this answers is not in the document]
raksharuia: Hmm, ok. But then we need to find a way to have atleast 2-3 digits after decimal displayed on the caption scale. I do think having simply 0.2 is not enough information.

perceval/rendering/pdisplay.py
raksharuia: Perhaps I did not ask correctly. I wanted to know the reason behind the dispatcher for `float` and `complex`. do we really need a special one just for them?

perceval/rendering/pdisplay.py
```diff
@@ -297,16 +397,9 @@ def pdisplay(o, output_format: Format = None, **opts):
...
-        opts_simple = {}
-        if "precision" in opts:
-            opts_simple["precision"] = opts["precision"]
-        if isinstance(o, (int, float)):
-            res = simple_float(o, **opts_simple)[1]
-        elif isinstance(o, complex):
-            res = simple_complex(o, **opts_simple)[1]
-        else:
-            raise RuntimeError("pdisplay not defined for type %s" % type(o))
+        return
```
ericbrts, line 407: I would have kept `raise RuntimeError("pdisplay not defined for type %s" % type(o))` instead of the return

perceval/rendering/pdisplay.py
```diff
@@ -284,13 +284,17 @@ def _get_sub_figure(ax: Axes3D, array: numpy.array, basis_name: list):
     ax.view_init(elev=30, azim=45)
 
 
-def pdisplay_tomography_chi(qpt: ProcessTomography, output_format: Format = Format.MPLOT, precision=1E-6):
-    if output_format != Format.MPLOT:
```
MarionQuandela: Can we add this parameter in the pdisplay method docstring pretty please ?

perceval/rendering/pdisplay.py
```diff
@@ -284,13 +284,17 @@ def _get_sub_figure(ax: Axes3D, array: numpy.array, basis_name: list):
...
-        raise TypeError("Tomography plot only support MPLOT")
+def pdisplay_tomography_chi(qpt: ProcessTomography, output_format: Format = Format.MPLOT, precision=1E-6,
+                            render_size=None):
```
MarionQuandela, line 288: Can we add the parameter `render_size` in the pdisplay method docstring pretty please ?

perceval/rendering/pdisplay.py
ericbrts: Moved the exception as NotImplementedError in a dispatched _pdisplay(object) function. So this comment is resolved

perceval/rendering/pdisplay.py
```diff
@@ -415,6 +415,8 @@ def pdisplay(o, output_format: Format = None, **opts):
                             level is drawn, others are "black boxes"
         - max_v (int): Maximum number of displayed values in distributions
         - sort (bool): if True, sorts a distribution (descending order) before displaying
+        - render_size: In SVG circuit/processor rendering, acts as a zoom factor (float)
```
MarionQuandela, line 418: This is the most beautiful documentation I've ever seen. Thanks for this piece of heaven

PR #311, opened 2023-11-10
Pcvl 585 clean up unnecessary operations on processor

raksharuia:


perceval/algorithm/tomography/tomography.py
```diff
@@ -107,81 +106,65 @@ class StateTomography(AAlgorithm):
...
-            input_state *= BasicState([m[1]])
-        input_distribution = self._source.generate_distribution(expected_input=input_state)
-        return input_distribution
+        self._nqubit = int(operator_processor.m / 2)  # nqubit
+        self._size_hilbert = 2 ** self._nqubit
+        self._gate_logical_perf = None
 
-    def _compute_probs(self, prep_state_indices, meas_pauli_basis_indices):
+    def _configure_processor(self, prep_state_indices, meas_pauli_basis_indices):
         # Adds preparation and measurement circuit to input processor (with the gate operation under study)
```
MarionQuandela, line 117: Can you use the docstrings convention ? Like that:
```
        """_summary_

        :param prep_state_indices: _description_
        :param meas_pauli_basis_indices: _description_
        :return: _description_
        """
```
And maybe in other places too ?

perceval/algorithm/tomography/tomography.py
```diff
@@ -107,81 +106,65 @@ class StateTomography(AAlgorithm):
...
         # circuit
-        # return: Output state probability distribution
-
-        pc = StatePreparationCircuit(prep_state_indices, self._nqubit)  # state preparation circuit object
-        mc = MeasurementCircuit(meas_pauli_basis_indices, self._nqubit)  # measurement basis circuit object
+        # returns the processor configured to perform state tomography experiment
 
-        p = Processor(self._backend, self._nqubit*2, self._source)
-        # A Processor with identical backend and source as the input
+        p = Processor(self._operator_processor.backend, self._nqubit * 2, self._operator_processor.source)
```
MarionQuandela, line 124: Maybe we should write somewhere that tomography will use the same backend as the processor ? And do we want to do that ? If the user uses Naive it will be a very long tomography no ?

perceval/algorithm/tomography/tomography.py
```diff
@@ -239,17 +219,16 @@ class ProcessTomography(AAlgorithm):
...
-        self._renormalization = renormalization
-        self._size_hilbert = 2 ** nqubit
-        self._qst = StateTomography(nqubit=self._nqubit, operator_processor=self._operator_processor,
-                                    post_process=self._post_process, renormalization=self._renormalization)
+        self._nqubit = int(operator_processor.m / 2)  # nqubit
+        self._size_hilbert = 2 ** self._nqubit
+        self._qst = StateTomography(operator_processor=self._operator_processor)
+
+        self.chi_normalized = None
+        self.chi_unnormalized = None
```
MarionQuandela, line 230: Why do we want to have the unormalized chi as a class member if we use it only in one method ?

perceval/algorithm/tomography/tomography.py
```diff
@@ -268,22 +247,14 @@ def _beta_as_matrix(self):
...
-                # only n changes but is mostly d**2
-                #
                 m, n = divmod(b, self._size_hilbert ** 2)
                 beta_matrix[a, b] = self._beta_tensor_elem(j, k, m, n, self._nqubit)
         return beta_matrix
 
     def _lambda_vector(self):
-        """
-        Computes the lambda vector of the operator
+        # Computes the lambda vector of the operator
```
MarionQuandela, line 255: This is not the correct way to do docstrings, but previous comment was.

tests/test_tomography.py
```diff
@@ -69,7 +69,7 @@ def test_fidelity_cnot_operator():
     cnot_p = catalog["klm cnot"].build_processor()
     cnot_op = np.array([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]], dtype='complex_')
 
-    cnot_fidelity = fidelity_op_process_tomography(cnot_op, cnot_p, nqubit=2)
+    cnot_fidelity = fidelity_op_process_tomography(cnot_op, cnot_p)
 
     assert cnot_fidelity == pytest.approx(1, 1e-3)  # computed fidelity is around 0.99967
```
MarionQuandela, line 74: If you rebase / merge from release/0.10.0 you will have a better fidelity since I fixed the KLM CNOT coefficients

MarionQuandela (changes requested):
Tomography notebook have to be upgraded with the new way of initializing tomographic object
Can you also reformat all the docstring that are not correct ? (with # and not """) we can do it together :)

perceval/algorithm/tomography/tomography.py
```diff
@@ -107,81 +106,65 @@ class StateTomography(AAlgorithm):
...
-        input_state = BasicState("|1,0>")
-        for _ in range(1, self._nqubit):
-            # setting the input state for the gate qubit modes
-            input_state *= BasicState("|1,0>")
-        for m in self._heralded_modes:
-            # setting the input for heralded modes of the given processor
-            input_state *= BasicState([m[1]])
-        input_distribution = self._source.generate_distribution(expected_input=input_state)
-        return input_distribution
+        self._nqubit = int(operator_processor.m / 2)  # nqubit
```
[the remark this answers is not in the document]
raksharuia: There Should be an error message as we are only supporting Dual Rail encoding so m should be double the number of Qubits

perceval/algorithm/tomography/tomography.py
raksharuia: I will update the docstring today :-)

perceval/algorithm/tomography/tomography.py
raksharuia: This was so that the method doesnot return multiple things but the class attribute gets updated once chi is computed; this will be given to a user only if they ask for it.

I can look up for cleaner/more logical solutions to achieve this.

perceval/algorithm/tomography/tomography.py
ericbrts: That's how Perceval works : using the stuff inside a processor in order to make a computation. So yes, we want to do that, and no need to inform the user we're doing so, I guess.
If a user wants to simulate with Naive, well... I just hope he's patient :)

perceval/algorithm/tomography/tomography.py
```diff
@@ -106,82 +45,68 @@ class StateTomography(AAlgorithm):
...
-        input_distribution = self._source.generate_distribution(expected_input=input_state)
-        return input_distribution
 
-    def _compute_probs(self, prep_state_indices, meas_pauli_basis_indices):
+    :param operator_processor: Gate under study
+    """
+    def __init__(self, operator_processor: AProcessor, **kwargs):
+        super().__init__(processor=operator_processor, **kwargs)
+        self._nqubit, odd_modes = divmod(operator_processor.m, 2)
+        if odd_modes:
```
ericbrts, line 54: Could you add a unit test showing that a 5 mode circuit does not trigger any computation?

perceval/algorithm/tomography/tomography.py
raksharuia: I mentioned the default config settings of Processor within `__init__()` of Tomography

tests/test_tomography.py
raksharuia: Could you do this while merging the PR?

MarionQuandela (approved):

PR #318, opened 2023-11-15
Pcvl 549 density matrix

melvinmathe:


perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,361 @@
...
+        for j, b_state in enumerate(B.reverse_index):
+            index = new_index[a_state*b_state]
+            perm[i*B.size+j, index] = 1
+
+    matrix = perm.T @ matrix @ perm
+
+    return DensityMatrix(matrix, new_index)
+
+
+class DensityMatrix:
```
ericbrts, line 113: General comment about this class: could you add typing hints in the method signatures?

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,361 @@
...
+
+        else:
+            return density_matrix_tensor_product(other, self)
+
+    def normalize(self):
+        """
+        Normalize the density matrix so that Trace(\rho) = 1
+        """
+
+        factor = self.mat.trace()
```
ericbrts, line 315: Even if not called in normalize, could you implement a `remove_low_magnitude_elements(threshold: float)` (or any other better name :) ) method to remove elements of magnitude lower than threshold?

tests/test_density_matrix.py
```diff
@@ -0,0 +1,106 @@
...
+
+    dm1 = DensityMatrix(svd1)
+    dm2 = DensityMatrix(svd2)
+
+    svd1_back = dm1.to_svd()
+    svd2_back = dm2.to_svd()
+
+    tensor_svd_back = (dm1*dm2).to_svd()
+    print(svd1)
+    print(svd1_back)
```
ericbrts, line 90: Can you remove these print statements?

tests/test_density_matrix.py
```diff
@@ -0,0 +1,106 @@
...
+    assert len(svd1_back) == len(svd1)
+    assert len(tensor_svd_back) ==  len(tensor_svd)
```
ericbrts, line 92: Is it possible to compare the svd contents (and not only their lengths)? If so, you can use the `assert_svd_close` function from file _test_utils.py

ericbrts (changes requested):

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+def array_to_statevector(vector, reverse_index):
+    """
+    translate an array in a StateVector
+    :param vector: an array
+    :param reverse_index: a list of BasicStates
+    """
+
+    sv = 0*reverse_index[0]
```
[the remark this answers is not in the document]
melvinmathe: I think I had a problem with the '|>' StateVector. When I do it this way, I immediately fix the right number of modes and I don't get this problem. I'll try to find an other way then.

tests/test_density_matrix.py
melvinmathe: That's what I did at the begining, but it doesn't work because the density matrix can be decomposed in different basis. So I'm still thinking about a way to do it properly. Do you have an idea ?

ericbrts: No, not really

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,361 @@
...
+                self.reverse_index = [None]*self._size
+                for key in index.keys():
+                    self.reverse_index[index[key]] = key
+            else:
+                raise ValueError("the index size does not match the matrix size")
+
+    def __getitem__(self, key):
+        """key must be a BasicState tuple"""
+        key1, key2 = key
+        if not isinstance(key1, BasicState) and isinstance(key2, BasicState):
```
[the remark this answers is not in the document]
ericbrts: Oh maybe you just need some parenthesis after the "not"

perceval/utils/density_matrix.py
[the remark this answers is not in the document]
ericbrts: You shouldn't have an issue. For me, the following code works:
```python
SV = pcvl.StateVector
FS = pcvl.BasicState
sv = SV()
for bs in [FS([1,1]), FS([1,2]), FS([2,1]), FS([2,2])]:
    sv += complex(0.9j)*bs

print(sv)
```

perceval/utils/density_matrix.py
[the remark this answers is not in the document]
melvinmathe: That was a mistake coming from a previous implementation

perceval/utils/density_matrix.py
melvinmathe: The probleme is that because I use a csr_array structure, removing items from the matrix will be inefficient (I'll probably get a warning) and I won't even gain memory...

ericbrts (approved):

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,363 @@
...
+                    new_term = (f"{self.mat[i, j]:.2f}*" +
+                                str(self.reverse_index[j]) +
+                                self._bra_str(self.reverse_index[i]) +
+                                "+")
+                    string += new_term
+
+        return string[:-1]
+
+    def __repr__(self):
+        str(self.mat.toarray())
```
MarionQuandela, line 347: It lack a `return` or else this method return nothing

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,363 @@
...
+
+        if not isinstance(other, DensityMatrix):
+            raise TypeError("You can only add a Density Matrix to a Density Matrix")
+
+        if not self._m == other._m:
+            raise ValueError("You can't add Density Matrices acting on different numbers of mode")
+
+        n = max(self._n_max, other.n_max)
+
+        if n == self.n_max:
```
MarionQuandela, line 274: Can't you refactor this bit of code ? Since it's almost the same.
Maybe by doing a method that will take the bigger matrix in first argument and the smaller as a second (so you will just have to permute the argument between self and other to make the right matrix)

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+
+        if isinstance(other, (int, float, complex)):
+            new_dm = copy(self)
+            new_dm.mat = other*new_dm.mat
+            return new_dm
+
+        return density_matrix_tensor_product(self, other)
+
+    def __rmul__(self, other):
```
MarionQuandela, line 299: Maybe this method should called the method `__mul__` but permuting arguments to avoid code duplication (maybe you have to put more test in the __mul__ ?)
```
def __rmul__(self, other):
    return __mul__(other, self)
```

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def __add__(self, other):
+        """
+        add two density Matrices together
+        """
+
+        if not isinstance(other, DensityMatrix):
+            raise TypeError("You can only add a Density Matrix to a Density Matrix")
+
+        if not self._m == other._m:
```
MarionQuandela, line 269: you can put instead
`if self._m != other._m:`

perceval/utils/density_matrix.py
```diff
@@ -0,0 +1,363 @@
...
+
+    matrix = perm.T @ matrix @ perm
+
+    return DensityMatrix(matrix, new_index)
+
+
+class DensityMatrix:
+    """
+    Density operator representing a mixed state
+    Does not support annotations
```
MarionQuandela, line 119: add a check in the __init__ to detect if there is annotation to throw an explicit message 
(method `has_annotations` allow this check)

MarionQuandela (changes requested):

perceval/utils/density_matrix.py
melvinmathe: In some cases, other will not be a density matrix, so if I call `__mul__` of an other class it will return something not implemented no ?

MarionQuandela: In this case ok

PR #330, opened 2023-12-06
break any n mode PERM into 2 mode PERM

raksharuia:


tests/test_permutation.py
```diff
@@ -69,3 +70,15 @@ def test_permutation_inverse():
     expected = [max(perm_vector)-i for i in perm_vector]
     expected.reverse()
     assert perm.perm_vector == expected
+
+
+@pytest.mark.parametrize("perm_list", [[2, 0, 1], [2, 3, 1, 0]])
+def test_n_mode_permutation_in_2_mode_perms(perm_list):
```
MarionQuandela, line 76: You can also add as a test that every components of this circuit is a 2 mode permutation
```
    for _, perm in new_circ:
        assert isinstance(perm, comp.PERM)
        assert perm.m == 2
```

MarionQuandela, line 76: I also try a random test a thousand times, it's working perfectly ! 😀 
```
        import random
        my_perm_list = list(range(15))
        random.shuffle(my_perm_list)
        n_mode_perm = comp.PERM(my_perm_list)
```

MarionQuandela (changes requested):
Congrats it works perfectly, I have very few comments, only some unnecessary variables, missing doc and test remarks ! :)

tests/test_permutation.py
raksharuia: Good suggestion :-) Implemented

raksharuia: I loved this test, i included it. Not running a 1000 times but only 10 haha

perceval/components/unitary_components.py
```diff
@@ -415,6 +415,23 @@ def apply(self, r, sv):
         })
         return nsv
 
+    def break_in_2_mode_perms(self):
+
+        perm_vec_req = self.perm_vector
```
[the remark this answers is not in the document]
raksharuia: Well, partly. I use it later in the code to keep checking the values and confirming mode swaps are going well so I thought of having it in a variable as it looked cleaner to me while working. If you prefer, I can change it to `self.perm_vector`

perceval/components/unitary_components.py
```diff
@@ -415,6 +415,23 @@ def apply(self, r, sv):
...
+        perm_len = len(perm_vec_req)
+
+        circ = Circuit(perm_len, name="PERM broken into 2 mode PERM")
```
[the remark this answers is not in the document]
raksharuia: Love the name! <3

perceval/components/unitary_components.py
```diff
@@ -415,6 +415,23 @@ def apply(self, r, sv):
         })
         return nsv
 
+    def break_in_2_mode_perms(self):
+
+        perm_vec_req = self.perm_vector
+        perm_len = len(perm_vec_req)
```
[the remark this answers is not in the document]
raksharuia: Technically calling the method `break_in_2_mode_perms()` would throw a different kind of error. Since the method returns a circuit, what the user has is not a `PERM` anymore. The method is defined in `PERM`  and not a `Circuit` so there would be an error.

Yes, I was planning to have a method like that in MPS, but you are correct - it would be more useful to have it in `Circuit`.

perceval/components/unitary_components.py
```diff
@@ -415,6 +415,32 @@ def apply(self, r, sv):
...
+        Breaks any n-mode PERM into an equivalent circuit with only 2 mode PERMs
+
+        :return: An equivalent Circuit with only 2 mode PERM components
+        """
+
+        perm_vec_req = self.perm_vector
+        perm_len = len(perm_vec_req)
+
+        if perm_len == 2:
+            print("length 2, returning myself")
```
ericbrts, line 429: Is this a debug trace? Can you remove it, please?

perceval/components/comp_utils.py
```diff
@@ -0,0 +1,47 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from .unitary_components import PERM
+from .linear_circuit import Circuit
+
+
+def decompose_perms(circuit: Circuit) -> Circuit:
```
ericbrts, line 34: Could you add a `merge` boolean parameter, which -if True- should not create additional hierarchical level in the circuit. You'd just pass this parameter on line 43 for it to work as expected:
```python
decomp_c.add(r, new_c, merge=merge)
```
I'd say the default value of merge should be True (?)

perceval/components/unitary_components.py
raksharuia: Oops, yes I missed to delete it. Thanks for catching it.

ericbrts (approved):

PR #335, opened 2023-12-22
Pcvl 550 pdisplay density matrices

melvinmathe:


perceval/rendering/pdisplay.py
```diff
@@ -320,10 +326,88 @@ def pdisplay_tomography_chi(qpt: ProcessTomography, output_format: Format = Form
     plt.show()
 
 
+def _complex_to_rgb(z: complex, cmap='hsv'):
```
ericbrts, line 329: Could you move your new utils fonctions to another python module (even a private one - i.e. prefix the py file name with an underscore)?

perceval/rendering/pdisplay.py
```diff
@@ -320,10 +326,88 @@ def pdisplay_tomography_chi(qpt: ProcessTomography, output_format: Format = Form
...
+
+
+def pdisplay_density_matrix(dm,
+                            output_format: Format = Format.MPLOT,
+                            color: bool = True,
+                            cmap='hsv'):
+
+    if output_format == Format.TEXT or output_format == Format.LATEX:
+        raise TypeError(f"DensityMatrix plot does not support {output_format}")
+    if color:
```
ericbrts, line 390: So `color` means you either display a colorful or a grayscale map. Ok. But, I don't understand what does the gray level of each pixel represents, in grayscale mode. The same but we lose the phase info?

Is `color` the correct name for this parameter?
At least add some docstring to explain what it does, please.

ericbrts (approved):

PR #338, opened 2024-01-08
random_unitary with params deprecated

raksharuia:


perceval/utils/matrix.py
```diff
@@ -145,19 +146,44 @@ def random_unitary(n: int, parameters: Optional[Union[np.ndarray,list]] = None)
...
         else:
             u = np.random.randn(n, n) + 1j*np.random.randn(n, n)
+            (q, r) = np.linalg.qr(u)
+            r_diag = np.sign(np.diagonal(np.real(r)))
+            n_u = np.zeros((n, n))
+            np.fill_diagonal(n_u, val=r_diag)
+            return Matrix(np.matmul(q, n_u))
+
+    @staticmethod
+    def parametrized_unitary(n: int, parameters: Union[np.ndarray,list] = None) -> MatrixN:
```
MarionQuandela, line 170: Can't we calculate n from the len of parameters ? And parameters can't be None right (default value) ?
So the method signature would be `def parametrized_unitary(parameters: Union[np.ndarray,list])`

tests/test_matrix.py
```diff
@@ -183,3 +184,26 @@ def test_genunitary():
...
+        assert issubclass(caught_warnings[0].category, DeprecationWarning)
+
+        # output Matrix
+        assert isinstance(M, pcvl.Matrix)
+        assert M.shape == (2, 2)
+        assert M.is_unitary()
+
+        # Testing the use of new method - parametrized_unitary()
+
+        PM = pcvl.Matrix.parametrized_unitary(2, np.arange(8))
```
MarionQuandela, line 205: This you can get it out of the with warnings statement

perceval/utils/matrix.py
```diff
@@ -145,19 +146,44 @@ def random_unitary(n: int, parameters: Optional[Union[np.ndarray,list]] = None)
...
         """
         if parameters is not None:
-            assert len(parameters) == 2*n**2, "parameters has not the right size: should be %d, and is %d" % (
-                2*n**2, len(parameters)
-            )
-            a = np.reshape(parameters[:n**2], (n, n))
-            b = np.reshape(parameters[n**2:], (n, n))
+            warnings.warn("use parametrized_unitary(parameters) instead to create a parametrized unitary "
+                          "matrix, version=0.11", DeprecationWarning)
+            assert len(parameters) == 2 * n ** 2, "parameters do not have the right size: should be %d, and is %d" % (
```
MarionQuandela, line 151: If parameters is not None, raise the deprecated warning and call parametrized_unitary no ?

MarionQuandela (changes requested):

perceval/utils/matrix.py
raksharuia: Yes, I missed to think of that and re-wrote the code. Ooops

perceval/utils/matrix.py
raksharuia: Yes, we can calculate n from the length of parameters however, it would involve a square root and I usually not prefer to have that as it would be a float and then you make it into an integer. Furthermore, with this signature the user can easily see that they are asking for a size 'n' matrix with '2*n^2' parameters.

Yes, in this case having `None` makes no sense. I will remove it. Thanks :)

perceval/utils/matrix.py
```diff
@@ -145,19 +146,35 @@ def random_unitary(n: int, parameters: Optional[Union[np.ndarray,list]] = None)
...
-            )
-            a = np.reshape(parameters[:n**2], (n, n))
-            b = np.reshape(parameters[n**2:], (n, n))
-            u = a + 1j * b
+            warnings.warn("use parametrized_unitary(parameters) instead to create a parametrized unitary "
+                          "matrix, version=0.11", DeprecationWarning)
+            return Matrix.parametrized_unitary(n, parameters)
         else:
             u = np.random.randn(n, n) + 1j*np.random.randn(n, n)
+            (q, r) = np.linalg.qr(u)
```
ericbrts, line 154: Could you factor the 4 lines that "make a matrix more unitary" in a private function / method, now that it is used in two different places?

perceval/utils/matrix.py
```diff
@@ -145,19 +146,35 @@ def random_unitary(n: int, parameters: Optional[Union[np.ndarray,list]] = None)
...
+            r_diag = np.sign(np.diagonal(np.real(r)))
+            n_u = np.zeros((n, n))
+            np.fill_diagonal(n_u, val=r_diag)
+            return Matrix(np.matmul(q, n_u))
+
+    @staticmethod
+    def parametrized_unitary(n: int, parameters: Union[np.ndarray,list]) -> MatrixN:
+        r"""static method generating a random unitary matrix
```
ericbrts, line 162: The method description needs to be updated

perceval/utils/matrix.py
```diff
@@ -145,19 +146,35 @@ def random_unitary(n: int, parameters: Optional[Union[np.ndarray,list]] = None)
         :return: a numeric Matrix
         """
         if parameters is not None:
-            assert len(parameters) == 2*n**2, "parameters has not the right size: should be %d, and is %d" % (
-                2*n**2, len(parameters)
-            )
-            a = np.reshape(parameters[:n**2], (n, n))
-            b = np.reshape(parameters[n**2:], (n, n))
-            u = a + 1j * b
+            warnings.warn("use parametrized_unitary(parameters) instead to create a parametrized unitary "
```
ericbrts, line 149: You request users to call `parametrized_unitary(parameters)` but shouldn't it be `parametrized_unitary(n, parameters)` ?

perceval/utils/matrix.py
raksharuia: oops

perceval/utils/matrix.py
raksharuia: yes

PR #281, opened 2023-10-02
DRAFT: [docs] sphinx multiversion

Its-Just-Nans:
It's a draft because it's missing the `index.html` which redirect to the correct version

Something like this to add in the CI

```jinja
<!DOCTYPE html>
<html>
  <head>
    <title>Redirecting to master branch</title>
    <meta charset="utf-8">
    <meta http-equiv="refresh" content="0; url=./{{latest}}">
    <link rel="canonical" href="{{latest}}">
  </head>
  <body>
  <p>Redirecting you to <a href="./{{latest}}">{{latest}}</a></p>
  </body>
</html>
````

PR #340, opened 2024-01-11
[docs] sphinx multiversion

Quandevla:
## Notes

- you can now close https://github.com/Quandela/Perceval/pull/281

You can remove tags we don't want with the regex `smv_tag_whitelist` like

```python
smv_tag_whitelist = r'^(?!(v0\.9\.3)|(v0\.9\.1))v\d+\.\d+\.\d+$'
```

To test which version sphinx is goinf to build you can do

```
cd docs && python -m sphinx_multiversion --dump-metadata source build/html
```


Quandevla:
To test this PR you can do

```
cd docs
make multiversion
cd build/html
python -m http.server
# go to webpage
```

Quandevla:
Note that we also can deploy the docs to github page with github actions ;)

MarionQuandela (changes requested):
So first congrats the multi version works, but not as is (the index main page try to redirect to v0.10.0.post14 that doesn't exists). But once we are into a version we can go from version to version. Then we don't need *all* doc I think, can we have a list of tags either we exclude or include in our doc generation ?

Quandevla:
I fix I have fixed the problem with redirection.

As explained, we can withelist versions (and kinda blacklist them)

```py
smv_tag_whitelist = r'^(?!(v0\.9\.3)|(v0\.9\.1))v\d+\.\d+\.\d+$'
```

Concerning the size, I think 50% of the generated docs are doctree files which are not useful for HTML (if I'm correct)


I also think that if you want to build the docs locally, we can use the `make html`

The `make multiversion` would be used only for automatic docs deployement with pipelines

PR #341, opened 2024-01-12
PCVL-607 bssamples serialization

ericbrts:


MarionQuandela (changes requested):
Can we add somewhere in UT an assert that the compress format is smaller than the not compressed format ?
For example in 
tests/test_serialization.py:test_compress
asset len(d_compressed['circuit']) < len(d_not_compressed['circuit'])

In this test I also notice than 
len(d_compressed['input_state']) > len(d_not_compressed['input_state'])
Is it normal ?

ericbrts:
> Can we add somewhere in UT an assert that the compress format is smaller than the not compressed format ? For example in tests/test_serialization.py:test_compress asset len(d_compressed['circuit']) < len(d_not_compressed['circuit'])
> 
> In this test I also notice than len(d_compressed['input_state']) > len(d_not_compressed['input_state']) Is it normal ?

Small amount of data get bigger when "compressed". That's consistent with the compression process we use, but default parameters (as to what types should be compressed when serialized) take that into account. Only larger data structure are compressed by default during serialization.

MarionQuandela (approved):

PR #344, opened 2024-01-12
fix get_results for asynchronous local job

raksharuia:


perceval/runtime/local_job.py
```diff
@@ -118,13 +119,16 @@ def cancel(self):
...
-        if self._result_mapping_function:
-            if 'results' in self._results:
-                self._results['results'] = self._result_mapping_function(self._results['results'],
-                                                                         **self._delta_parameters['mapping'])
-            elif 'results_list' in self._results:
-                for res in self._results["results_list"]:
-                    res["results"] = self._result_mapping_function(res['results'], **self._delta_parameters['mapping'])
-            else:
-                raise KeyError("Cannot find either 'result' or 'results_list' in self._results")
+        if not self._is_res_mapped:
```
ericbrts, line 122: I don't think you need a new variable to remember that a result was converted. Actually, the flow of the function is, if you have a `self._result_mapping_function` it means you have to convert (or map) results.

So I'd say, at the end of the `if self._result_mapping_function:` block, you can just remove the mapping function with a `self._result_mapping_function = None` and get rid of your new variable.

PR #346, opened 2024-01-19
Pcvl 551 evolve density matrix

melvinmathe:


perceval/simulators/simulator.py
```diff
@@ -425,7 +430,7 @@ def evolve(self, input_state: Union[BasicState, StateVector]) -> StateVector:
         result_sv.normalize()
         return self._post_select_on_statevector(result_sv)
 
-    def evolve_svd(self, svd: Union[SVDistribution, StateVector, BasicState], progress_callback: Optional[Callable] = None) -> SVDistribution:
+    def evolve_svd(self, svd: Union[SVDistribution, StateVector, BasicState], progress_callback: Optional[Callable] = None):
```
ericbrts, line 433: `...) -> dict:`
C'est toujours mieux que de ne rien mettre en type de retour

perceval/utils/density_matrix_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+from perceval.utils.statevector import StateVector
+from typing import Union
+import numpy as np
+from scipy.sparse import sparray, csr_array
+
+
+def extract_upper_triangle(csr_matrix: csr_array) -> csr_array:
+    """
+    extract the upper half matrix S of a hermitian matrix rho such that S + S_dagger = rho
```
ericbrts, line 38: Why does the csr_matrix have to be hermitian? I think in this docstring you're only describing *your* use case of this function. But this function is just extracting the top right half of any square sparse matrix, isn't it?
If so the docstring should just say that and nothing more. 

If you want to tell other devs that you're extracting the triangle of an hermitian matrix because it simplifies the computation, given the S + Sdagger = rho rule, please do so, but not here (I guess that would be in the `Simulator.evolve_density_matrix` method)

perceval/utils/density_matrix.py
```diff
@@ -27,7 +27,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from perceval.utils.statevector import StateVector, SVDistribution, BasicState, max_photon_state_iterator
+from perceval.utils.statevector import SVDistribution, BasicState, max_photon_state_iterator
+from perceval.utils.density_matrix_utils import *
```
ericbrts, line 31: Please avoid as much as possible `import *`

ericbrts (approved):

PR #352, opened 2024-01-26
Pcvl 638 probs density matrix

melvinmathe:


ericbrts (approved):

perceval/simulators/simulator.py
```diff
@@ -390,6 +390,32 @@ def probs_svd(self, input_dist: SVDistribution, progress_callback: Optional[Call
...
+        """
+        gives the output probability distribution, after evolving some density matrix through the simulator
+        :param dm: the input DensityMatrix
+        """
+        if not isinstance(dm, DensityMatrix):
+            raise TypeError(f"dm must be a DensityMatrix object, {type(dm)} was given")
+
+        input_list = self._get_density_matrix_input_list(dm)
+        u_evolve = self._construct_evolve_operator(input_list, dm)
+        u_evolve_in_row = csr_array(u_evolve)
```
raksharuia, line 403: Simply to clarify, you use `csc_array` for evolve and `csr_array` for probs - any specific reason?

Also, `dok_array` is used in `DensityMatrix` (I see that it seems easier to build from scratch). Perhaps some documentation/comment would be nice to help remember their usage for posterity.

perceval/simulators/simulator.py
```diff
@@ -474,19 +500,26 @@ def evolve_density_matrix(self, dm: DensityMatrix) -> DensityMatrix:
         """
         if not isinstance(dm, DensityMatrix):
             raise TypeError(f"dm must be of DensityMatrix type, {type(dm)} was given")
-        size = dm.mat.shape[0]
 
         # Establishing te set of FockState to evolve
```
raksharuia, line 504: typo in this comment :-D

perceval/simulators/simulator.py
```diff
@@ -501,8 +534,15 @@ def evolve_density_matrix(self, dm: DensityMatrix) -> DensityMatrix:
...
+        return u_evolve
 
-        inter_matrix = u_evolve @ extract_upper_triangle(dm.mat) @ u_evolve.T.conj()
-        out_matrix = inter_matrix + inter_matrix.T.conjugate(copy=False)
-
-        return DensityMatrix(out_matrix, index=dm.index)
+    @staticmethod
+    def _get_density_matrix_input_list(dm: DensityMatrix) -> list:
+        """
+        get the input_list necessary to density_matrix evolution
```
raksharuia, line 542: `input_list` is used in both evolve and probs of density matrix. Perhaps a different way of documenting/classifying it?

raksharuia (commented):
Good work :-) I simply have tiny comments and questions

perceval/simulators/simulator.py
melvinmathe: csc_arrays can be constructed by columns (and then we can iterate on columns), and that's what I do for the evolve operator because each column correspond to the output of one Fockstate. However in probs, I need to iterate on the rows so I convert the operator into this format first.
dok_array are less efficient but I can add values anywhere in the matrix easily, so that's what I do for very sparse matrix, such as basis change matrix, etc
I added some comments.

raksharuia (approved):

PR #354, opened 2024-02-02
PCVL-655 NoiseModel

ericbrts:


perceval/utils/noise_model.py
```diff
@@ -0,0 +1,109 @@
...
+        return {k: v.get() for k, v in self._params.items() if not v.is_default}
+
+    def __eq__(self, other):
+        for field in self._params:
+            if self._params[field] != other._params[field]:
+                return False
+        return True
+
+
+if __name__ == "__main__":
```
raksharuia, line 95: You have this segment to debug while developing by running as a standalone script, isn't it?

When I tried running the `noise_model.py` as a standalone script, it did not work.

raksharuia (commented):
Everything looks good and the tests run well on my local machine. I also did some random nm object creation with params and error messages appeared correctly. Good work :-)

* I like the idea of ValidatedParams class - just one question : Do we plan to have more than bool and float as param types?
* any particular reason to choose the name of the file _validated_fields?

perceval/utils/noise_model.py
ericbrts: You're right, I forgot to clean up my file before committing.

ericbrts:
>     * I like the idea of ValidatedParams class - just one question : Do we plan to have more than bool and float as param types?

Yes, but I'd like to add them only when we need them.

>     * any particular reason to choose the name of the file _validated_fields?

I changed the class names from  `SomethingField` to `SomethingParam` and forgot to change the file name.

PR #356, opened 2024-02-08
Pcvl 644 loss operator density matrix

melvinmathe:


perceval/utils/density_matrix.py
```diff
@@ -363,6 +364,48 @@ def sample(self, count: int = 1):
...
+        """
+
+        listed_fs = list(fockstate)
+        if listed_fs[m] <= n_photon:
+            listed_fs[m] = 0
+        else:
+            listed_fs[m] -= n_photon
+        return BasicState(listed_fs)
+
+    def _construct_loss_operators(self, mode: int, p:float):
```
raksharuia, line 380: I am a little confused with how the loss is modeled here; the parameter `p` here comes from `probs` - does it mean that the user inputs some sort of loss probability?

tests/test_density_matrix.py
```diff
@@ -156,3 +158,20 @@ def test_remove_low_amplitude():
...
+    assert dm.mat.trace() == pytest.approx(1)
+
+    for k in range(6):
+        assert dm[BasicState([k]), BasicState([k])] == pytest.approx(math.comb(5, k) * (1/2)**5)
+
+    sv = BasicState([2, 1, 0]) + BasicState([1, 3, 2]) + BasicState([0, 0, 1])
+    dm = DensityMatrix.from_svd(sv)
+    dm.apply_loss([0, 1], .2)
+
+    assert dm.mat.trace() == pytest.approx(1)
```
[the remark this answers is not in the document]
melvinmathe: I agree. It's not enough of a test but I thought it was necessary at first to be sure that the computation was giving back something normal. The tests in the loop above are more precise because they check the number of loosed photon is actually following a binomial probability law (which is what is expected)

perceval/utils/density_matrix.py
melvinmathe: yes
I agree that prob might not be the best name.

tests/test_density_matrix.py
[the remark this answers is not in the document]
raksharuia: could a test like that be added for the second case too?

perceval/utils/density_matrix.py
```diff
@@ -363,6 +364,63 @@ def sample(self, count: int = 1):
...
+                    #  Binomial probability, (comes from the simplification of the beam splitter action)
+                    operators[n_photon_loss][result_idx, idx] += sqrt(comb(n_photon, n_photon_loss) *
+                                                                      (1-p)**(n_photon-n_photon_loss) *
+                                                                      p**n_photon_loss)
+        return operators
+
+    @dispatch(int, float)
+    def apply_loss(self, mode: int, prob: float):
+        """
+        Apply a loss on some mode according to some probability of loosing a photon
```
ericbrts, line 403: Typo "losing", I think

perceval/utils/density_matrix.py
```diff
@@ -363,6 +364,63 @@ def sample(self, count: int = 1):
...
+        """
+        matrix_after_loss = csr_array(self.shape, dtype=complex)
+        for operator in self._construct_loss_operators(mode, prob):
+            matrix_after_loss += operator @ self.mat @ operator.T
+        self.mat = matrix_after_loss
+
+    @dispatch(list, float)
+    def apply_loss(self, modes: list, prob: float):
+        """
+        Apply a loss on some modes according to some probability of loosing a photon
```
ericbrts, line 416: losing ?

ericbrts (approved):

raksharuia (approved):

PR #361, opened 2024-02-15
Pcvl 647 optimize allstate iterator

melvinmathe:


perceval/backends/_slos.py
```diff
@@ -197,7 +198,7 @@ def prob_distribution(self) -> BSDistribution:
         c = abs(c) ** 2 / istate.prodnfact()
         xq.all_prob_normalize_output(c, self._fsas[istate.n])
         bsd = BSDistribution()
-        for output_state, probability in zip(allstate_iterator(self._input_state, self._mask), c):
+        for output_state, probability in zip(self._get_iterator(self._input_state, self._mask), c):
```
raksharuia, line 201: would this change be only in _slos?

melvinmathe: Nope. The _get_iterator method comes from the AProbampliBackend class and can be used in any backend

raksharuia (approved):

perceval/backends/_abstract_backends.py
```diff
@@ -70,6 +70,19 @@ def samples(self, count: int):
...
+
+    def __init__(self):
+        super().__init__()
+        self._cache_iterator = dict()
+
+    def _get_iterator(self, input_state: BasicState, mask=None):
+        n_photons = input_state.n
+
+        if n_photons not in self._cache_iterator.keys():
+            self._cache_iterator[n_photons] = list(allstate_iterator(input_state, mask))
```
ericbrts, line 82: The result of the `allstate_iterator` changes given **n** (number of photons), **m** (number of modes) and the mask.
Indexing the `allstate_iterator` cache only on **n** is doable, only if you invalidate the whole cache (reset it to an empty dict) when **m** (m can change when a new circuit is set) or a mask changes.

Alternatively, you could index on a tuple containing (m, n, mask) - making it less readable maybe.

perceval/backends/_abstract_backends.py
```diff
@@ -70,6 +70,19 @@ def samples(self, count: int):
 
 
 class AProbAmpliBackend(ABackend):
+
+    def __init__(self):
+        super().__init__()
+        self._cache_iterator = dict()
```
ericbrts, line 76: Could you add a `clear_iterator_cache()` method to free the memory if needed?

melvinmathe: Yes, I'm on it.

perceval/backends/_abstract_backends.py
melvinmathe: I think I'll rather invalidate the cache If you think it's ok...

perceval/backends/_abstract_backends.py
```diff
@@ -70,6 +70,28 @@ def samples(self, count: int):
...
+
+    def _get_iterator(self, input_state: BasicState, mask=None):
+        n_photons = input_state.n
+
+        if n_photons not in self._cache_iterator.keys():
+            self._cache_iterator[n_photons] = list(allstate_iterator(input_state, mask))
+
+        return self._cache_iterator[n_photons]
+
+    def _reset(self):
```
ericbrts, line 86: Ok but just make it public and with a more specific name, like `clear_iterator_cache()`.

You don't **have to** use SLOS method names

tests/test_backends.py
```diff
@@ -277,3 +277,14 @@ def test_backend_mps_n_mode_perm_decomp():
     check_output_distribution(backend, BasicState("|1,0,1>"),
                               {BasicState("|0,2,0>"): 0.5,
                                BasicState("|0,0,2>"): 0.5})
+
+
+def test_probampli_iterator_cache():
+    backends = [SLOSBackend(), MPSBackend(), NaiveBackend()]
```
ericbrts, line 283: We usually do a parametrized auto test with `@pytest.mark.parametrize` instead of looping over different values (see test_backend_cnot for an example).

You can `@pytest.mark.parametrize(backend_type, (SLOSBackend, MPSBackend, NaiveBackend))` and then call
`b = backend_type()` to instanciate

perceval/backends/_abstract_backends.py
melvinmathe: ok

tests/test_backends.py
melvinmathe: ok

ericbrts (approved):

PR #362, opened 2024-02-15
PCVL-641 Fix probs_to_sample_count conversion for a low number of samples

ericbrts:


perceval/utils/conversion.py
```diff
@@ -59,13 +59,18 @@ def samples_to_probs(sample_list: BSSamples) -> BSDistribution:
 
 def probs_to_sample_count(probs: BSDistribution, count: int = None, **kwargs) -> BSCount:
     count = _deduce_count(count, **kwargs)
+    if count < 1:
+        return BSCount()
     perturbed_dist = {state: max(prob + np.random.normal(scale=(prob * (1 - prob) / count) ** .5), 0)
                       for state, prob in probs.items()}
```
melvinmathe, line 65: It's just my ignorance but why do you add some randomness to the distributon. Will not it appear from the sampling automatically ?

perceval/utils/conversion.py
```diff
@@ -59,13 +59,18 @@ def samples_to_probs(sample_list: BSSamples) -> BSDistribution:
...
     if prob_sum == 0:
         return samples_to_sample_count(probs_to_samples(probs, count))
     fac = 1 / prob_sum
     perturbed_dist = {key: fac * prob for key, prob in perturbed_dist.items()}  # Renormalisation
+    if max(perturbed_dist.values()) * count < 1:
+        return samples_to_sample_count(probs_to_samples(probs, count))
+
     results = BSCount()
     for state in perturbed_dist:
         results.add(state, round(perturbed_dist[state] * count))
```
melvinmathe, line 76: Oh ok that's why...

melvinmathe (approved):
I made the test and it works !

perceval/utils/conversion.py
Benoit-F-Q: I still do not get it

Benoit-F-Q:
I don't understand why 'results' will be empty. I think it should contains all keys from prob.items(), with value 0

perceval/utils/conversion.py
ericbrts: This function converts an exact probability distribution to a sample count, _as if it was really sampled_.
The perturbation in the input exact distribution simulates this sampling.

raksharuia: @ericbrts, just one question - I fail to see how the condition (L71) `max(perturbed_dist.values()) * count < 1` removes the exception and quantifies low sample count.

Benoit-F-Q: This one I can answer : for every key s.t. ```perturbated_dist.value * count < 1``` you will get a 0 simulated sample count.

Thus if max(XXX)*count < 1, there will be no entry at all.

Benoit-F-Q: Anyway, after one night thinking of it, I'd rather replace ```results[random.choice(list(results.keys()))] += diff``` with ```results[random.choice(list(probs.keys()))] += diff```.

probs is not empty (checked by ```if prob_sum == 0```) and this would avoid real sampling when we said we would simulate it.

But this doesn't really change the beahavior, so this remark is "just a thought".

Benoit-F-Q: (and I hope that at some time someone will say "Don't worry Benoit, the random seed is freezed during tests", because debugging a non-reproductable unit test looks like hazing)

ericbrts: The seed isn't fixed during auto tests - and we've got (rare) issues because of that.

Benoit-F-Q (approved):
Well done 🥇

PR #364, opened 2024-02-15
PCVL 663 quickfix

melvinmathe:


perceval/simulators/simulator.py
```diff
@@ -408,7 +408,7 @@ def probs_density_matrix(self, dm: DensityMatrix) -> dict:
         for row_idx, fs in enumerate(dm.reverse_index):
 
             vec = u_evolve_in_row[[row_idx]]
-            prob = abs(vec @ dm.mat @ vec.conj().T)
+            prob = abs((vec @ dm.mat @ vec.conj().T)[0, 0])
```
raksharuia, line 411: there seems to be no problem with the fix - but I am missing some context here.

melvinmathe: The goal is just to get a float instead of a (1,1) sparse matrix in the output BSD

PR #368, opened 2024-02-20
Pcvl 652 dm doc

melvinmathe:


perceval/utils/density_matrix.py
```diff
@@ -500,27 +503,27 @@ def _construct_loss_operators(self, mode: int, p: float):
...
         Apply a loss on some mode according to some probability of losing a photon
         Everything works like if the mode was connected to some virtual mode with a beam splitter of reflectivity prob
-        :param mode: the mode were you want to simulate a loss
+
+        :param modes: the mode were you want to simulate a loss
         :param prob: the probability to lose a photon
         """
+        self._apply_loss(modes, prob)
+
+    @dispatch(int, float)
```
ericbrts, line 516: In order to get the docstring automatically retrieved by Sphinx, you might want to either:

-  try single dispatch (with `@functools.singledispatch`, see: https://docs.python.org/3/library/functools.html
- gather both signatures to only one, and if the user passes an int, just turn it to a single element list, in order to have only one working implementation of your algorithm (that would expect a List[int] input parameter)

docs/source/states.rst
```diff
@@ -204,3 +204,22 @@ results in the mixed state ``1/2|0,1><0,1|+1/4(1/sqrt(2)*|1,0>+1/sqrt(2)*|0,1>)(
...
+class instance to the method ``from_svd()``.
+
+Many of the above methods extends to Density Matrices, such as ``sample`` and ``measure``. This class is also suited for non unitary evolution of the state.
+For example, the ``apply_loss`` method is already build-in to simulate the probabilistic loss of photons. It is also handled by the simulator for unitary evolutions
+with the method ``evolve_density_matrix``.
+
+The ``pdisplay()`` function also allows to easily see the coherence of your state by highlighting the non diagonal coefficients. For exemple, here is the density resulting matrix
+when we try to create a bell state with a Post-Selected CNOT, with a noisy source:
+
+.. figure:: _static/img/bell.svg
```
ericbrts, line 223: Should users know what it would look like for a perfect source?
If not, do you think it might be a good idea to display perfect vs noisy here?

ericbrts (approved):

PR #371, opened 2024-02-23
PCVL-656 NoiseModel in local Processor

ericbrts:


perceval/components/processor.py
```diff
@@ -53,27 +53,57 @@ class Processor(AProcessor):
...
+        self._init_circuit(m_circuit)
+        self._init_noise(source, noise)
+        self.name = name
+        self._inputs_map: Union[SVDistribution, None] = None
+        self._simulator = None
 
+    def _init_noise(self, source: Source, noise: NoiseModel):
+        # Default = perfect simulation
+        self._phase_quantization = 0
+        self._source = Source()
```
MarionQuandela, line 73: You can initialize _source to None, and then at the end of the method do:
```
if self_source is None:
    self._source = Source()
```
instead of initializing an object you may not need ?

perceval/components/processor.py
```diff
@@ -53,27 +53,57 @@ class Processor(AProcessor):
...
+        elif noise is not None:
+            self.noise = noise
+
+    @AProcessor.noise.setter
+    def noise(self, nm):
+        super(Processor, type(self)).noise.fset(self, nm)
+        self._source = Source.from_noise_model(nm)
+        self._phase_quantization = nm.phase_imprecision
+        if isinstance(self._input_state, BasicState):
+            self._generate_noisy_input()
```
MarionQuandela, line 92: This is cool to have thought about handle the source, but don't we need to do something about phase imprecision also ? Like rebuilding the circuit ?

tests/test_processor.py
```diff
@@ -222,3 +246,19 @@ def test_add_remove_ports():
     for i in range(6):
         assert processor.get_input_port(i) is None
         assert processor.get_output_port(i) is None
+
+
+def test_phase_quantization():
```
MarionQuandela, line 251: This is an indirect test, could we also test the phase values of p2 in this test ?

MarionQuandela (changes requested):
If I'm not mistaken there is not a test with both source noise parameters and phase quantization noise ? Can we add one pretty please <3 ?
Also changing the noise model seems to have side effects ? Maybe we can add test around that and explicitly put in documentation what's working and what's not working

tests/test_processor.py
```diff
@@ -222,3 +246,19 @@ def test_add_remove_ports():
...
+    p1 = Processor("SLOS", catalog["mzi phase first"].build_circuit(phi_a=0.596898191919898198,
+                                                                    phi_b=0.16561561651616), noise=nm)
+    p2 = Processor("SLOS", catalog["mzi phase first"].build_circuit(phi_a=0.6,
+                                                                    phi_b=0.2))
+
+    p0.with_input(BasicState([1, 1]))
+    p1.with_input(BasicState([1, 1]))
+    p2.with_input(BasicState([1, 1]))
+    assert p0.probs()["results"] != pytest.approx(p1.probs()["results"])
+    assert p1.probs()["results"] == pytest.approx(p2.probs()["results"])
```
MarionQuandela, line 264: This is not working
```
    p1.noise = NoiseModel()
    assert p0.probs()["results"] == pytest.approx(p1.probs()["results"])
```
This is working
```
    p0.noise = nm
    assert p0.probs()["results"] == pytest.approx(p1.probs()["results"])
```
I understand that we cannot "undo" the calculation of the phases for know but we either need to support it or to put a warning when going from phase imprecision to no phase imprecision.

perceval/utils/noise_model.py
```diff
@@ -38,7 +38,7 @@ class NoiseModel:
 
     :param brightness: first lens brightness of a quantum dot
     :param indistinguishability: chance two photons are indistinguishable
-    :param g2: chance two photons are emitted instead of one (g²(0) value)
+    :param g2: g²(0) - second order intensity autocorrelation at zero time delay
```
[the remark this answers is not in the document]
ericbrts: "Chance" isn't the correct word, as g² isn't a probability. As I couldn't phrase it well, I got back to the definition from the `Source` class

perceval/components/processor.py
ericbrts: The method `linear_circuit` is handling the phase imprecision parameter (that's why I've overridden this method in `Processor`). I've added a docstring to this implementation so that it's clearer.

tests/test_processor.py
ericbrts: I only want to test that _manually reducing the precision of phase values_ has exactly the same effect as _using the `phase_imprecision` noise parameter_. I do not care about the exact probability results in this test - as this would be more like testing the simulation back-end than the noise handling of the `Processor` (so I really think this does not belong to this test).

tests/test_processor.py
ericbrts: Thanks for having caught that. I've copied the circuit that's used in the simulation, so that we still know the original phase values.

perceval/utils/noise_model.py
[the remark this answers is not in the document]
MarionQuandela: I found this post: https://physics.stackexchange.com/questions/139833/what-is-g2-in-the-context-of-quantum-optics-and-how-is-it-calculated
What about:
g²(0) - second order intensity autocorrelation at zero time delay. Meaning how often we detect two photons at the same time (detector-wise)

PR #379, opened 2024-04-02
Pcvl 686 explicit cast to integer

Benoit-F-Q:
I didn't reproduce the `Raise` with :
```
def input_type(key: int):
    print("input conversion: {}".format(key))

def return_type(**kwargs) -> int:
    return kwargs.get("key", None)

value = 1e6
input_type(key = value)
print("output conversion: {}".format(return_type(key=value)))
```

Where does the check come from? Protobuf? Somewhere else?

Shouldn't we fix in a more generic way to prevents similar issues?

### DIFF
--- a/perceval/algorithm/abstract_algorithm.py
+++ b/perceval/algorithm/abstract_algorithm.py
@@ -44,7 +44,7 @@ class AAlgorithm:
         if not self._check_compatibility():
             raise RuntimeError("Processor and algorithm are not compatible")
         self.default_job_name = None
-        self._max_shots = kwargs.get(self._MAX_SHOTS_NAMED_PARAM)
+        self._max_shots = int(kwargs.get(self._MAX_SHOTS_NAMED_PARAM))
 
     def _check_compatibility(self) -> bool:
         # if self._processor.is_remote:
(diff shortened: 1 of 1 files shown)


perceval/algorithm/abstract_algorithm.py
```diff
@@ -44,7 +44,10 @@ def __init__(self, processor: AProcessor, **kwargs):
         if not self._check_compatibility():
             raise RuntimeError("Processor and algorithm are not compatible")
         self.default_job_name = None
-        self._max_shots = kwargs.get(self._MAX_SHOTS_NAMED_PARAM)
+        try:
+            self._max_shots = int(kwargs.get(self._MAX_SHOTS_NAMED_PARAM))
+        except:
+            self._max_shots = None
```
ericbrts, line 50: Here, you should raise an exception. We require users to feed a shots threshold to algorithms, else they could burn their shots (credits) without knowing about it.

Benoit-F-Q: This is already checked at the start of this method, for remote jobs only.

Many TU are designed without max_shots

PR #380, opened 2024-04-02
PCVL-694 noisy improvements

ericbrts:


perceval/simulators/simulator.py
```diff
@@ -300,31 +300,40 @@ def _probs_svd_generic(self, input_dist, p_threshold, progress_callback: Optiona
...
         for sv, prob in input_dist.items():
             if max(sv.n) >= self._min_detected_photons:
                 decomposed_input.append(
                     (prob, sv[0].separate_state(keep_annotations=False))
                 )
             else:
                 self._physical_perf -= prob
-        input_set = set([state for s in decomposed_input for state in s[1]])
+
+        """Create a cache with strong simulation of all unique input"""
```
MarionQuandela, line 325: Why comments with """ and not # ?

MarionQuandela, line 325: Why comments with """ and not # ?

perceval/simulators/simulator.py
```diff
@@ -300,31 +300,40 @@ def _probs_svd_generic(self, input_dist, p_threshold, progress_callback: Optiona
...
+        for idx, state in enumerate(input_set):
             self._backend.set_input_state(state)
             cache[state] = self._backend.prob_distribution()
+            if progress_callback and idx % 10 == 0:
+                progress = (idx + 1) / len_input_set * 0.5  # From 0. to 0.5
+                exec_request = progress_callback(progress, 'compute probability distributions')
+                if exec_request is not None and 'cancel_requested' in exec_request and exec_request['cancel_requested']:
+                    raise RuntimeError("Cancel requested")
 
         """Reconstruct output probability distribution"""
```
MarionQuandela, line 338: Why comments with """ and not # ?

perceval/utils/statevector.py
```diff
@@ -239,6 +239,22 @@ def m(self):
...
+        """
+        if len(svd1) == 0:
+            return svd2
+        new_dist = SVDistribution()
+        for sv1, proba1 in svd1.items():
+            for sv2, proba2 in svd2.items():
+                if proba1 * proba2 < prob_threshold:
+                    continue
+                sv = sv1 * sv2
+                new_dist[sv] += proba1 * proba2
```
MarionQuandela, line 255: new_dist[sv] doesn't exists, no ? So you can't do += ? Or is it handle in the __getitem__method ?

perceval/components/source.py
```diff
@@ -181,18 +181,21 @@ def probability_distribution(self, nphotons: int = 1) -> SVDistribution:
...
         :param expected_input: Expected input BasicState
-        The properties of the source will alter the input state. A perfect source always delivers the expected state as
-        an input. Imperfect ones won't.
+            The properties of the source will alter the input state. A perfect source always delivers the expected state
+            as an input. Imperfect ones won't.
+        :param simplify: Simplify the distribution by anonymizing photon annotations (can be time-consuming for larger
+            distributions). Default is False.
         """
         dist = SVDistribution()
+        prob_threshold = global_params['min_p']
```
MarionQuandela, line 196: What does min_p means ? minimum probability ? and why is it so low (1e-16) ?

perceval/utils/statevector.py
```diff
@@ -239,6 +239,22 @@ def m(self):
     def n_max(self):
         return self._n_max
 
+    @staticmethod
+    def tensor_product(svd1, svd2, prob_threshold: float = 0):
```
MarionQuandela, line 243: This is a good method of simplification ! Are you sure about the default value ? We don't want 1e-6/1e-16 instead of zero ? I don't see any UT for this method, can you implement one real quick ?

perceval/simulators/simulator.py
```diff
@@ -300,31 +300,40 @@ def _probs_svd_generic(self, input_dist, p_threshold, progress_callback: Optiona
...
+
+        """Create a cache with strong simulation of all unique input"""
         cache = {}
-        for state in input_set:
+        input_set = set([state for s in decomposed_input for state in s[1]])
+        len_input_set = len(input_set)
+        for idx, state in enumerate(input_set):
             self._backend.set_input_state(state)
             cache[state] = self._backend.prob_distribution()
+            if progress_callback and idx % 10 == 0:
```
MarionQuandela, line 332: Did you have time to check this part or do you want me to do it ?

perceval/utils/statevector.py
```diff
@@ -239,6 +239,22 @@ def m(self):
...
+        if len(svd1) == 0:
+            return svd2
+        new_dist = SVDistribution()
+        for sv1, proba1 in svd1.items():
+            for sv2, proba2 in svd2.items():
+                if proba1 * proba2 < prob_threshold:
+                    continue
+                sv = sv1 * sv2
+                new_dist[sv] += proba1 * proba2
+        return new_dist
```
Benoit-F-Q, line 256: Ne faut-il pas renormaliser la SVD, si on a filtré des SV ?

perceval/components/source.py
ericbrts: Yes, it's the minimum probability we consider Perceval-wide. It's low because we don't want to scrap too many "low contribution" from states. But in several algorithms (e.g. the noisy algo in Simulator), we're not using this global value as threshold.

perceval/simulators/simulator.py
ericbrts: Yes, I'm checking this a lot on the cloud worker side

perceval/utils/statevector.py
ericbrts: That's right, but I didn't normalize here to avoid unnecessary computations. I've normalized after the source has added all the noise in the distribution.

PR #390, opened 2024-05-16
PCVL-703 Naive backend with Gurvitz estimator

Benoit-F-Q:


perceval/backends/_naive.py
```diff
@@ -39,6 +39,10 @@ class NaiveBackend(AProbAmpliBackend):
     """Naive algorithm, no clever calculation path, does not cache anything,
        recompute all states on the fly"""
 
+    def __init__(self, permanentor = lambda M : xq.permanent_cx(M, n_threads=1)):
```
ericbrts, line 42: I'm not a fan of letting any Naive user inject a way to compute permanent in the main interface of the class.

I'd suggest a simpler approach: a `_compute_permanent()` method which is overriden by the approximate naive back-end

ericbrts (changes requested):

PR #394, opened 2024-05-17
Pcvl-695 create perceval logging

MarionQuandela:


perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
...
+        if "channel" in record.__dict__:
+            if record.levelno < self[_CHANNELS][record.channel]["level"]:
+                return False
+        return True
+
+
+CONFIG = LoggerConfig()
+
+
+class ExqaliburLogger():
```
ericbrts, line 114: As you created two proxy classes with the same interface, you can create an interface `ABC` class defining all the abstract methods you expect

perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
...
+            print(f"starting to write logs in {log_path}")
+            exqalibur_logging.enable_file()
+        else:
+            exqalibur_logging.disable_file()
+
+        exqalibur_logging.enable_console()
+
+        if _CHANNELS in CONFIG:
+            for name, value in CONFIG[_CHANNELS].items():
+                exqalibur_logging.set_level(_LEVEL_CONVERTER[value["level"]], _CHANNEL_CONVERTER[name])
```
ericbrts, line 134: What happens if the config contains invalid values?

perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
...
+    def initialize():
+        persistent_data = PersistentData()
+        log_path = ExqaliburLogger.get_log_file_path()
+        if persistent_data.is_writable():
+            exqalibur_logging.initialize(log_path)
+        else:
+            exqalibur_logging.initialize()
+
+        if _ENABLE_FILE in CONFIG and CONFIG[_ENABLE_FILE]:
+            print(f"starting to write logs in {log_path}")
```
ericbrts, line 125: I'd remove this print

perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
...
+    def error(self, msg: str, channel: str = "user", exc_info=None):
+        self._logger.error(f"[error] {msg}", exc_info=exc_info, extra={
+                           "channel": channel})
+
+    def critical(self, msg: str, channel: str = "user", exc_info=None):
+        self._logger.critical(f"[critical] {msg}", exc_info=exc_info, extra={
+                              "channel": channel})
+
+
+ExqaliburLogger.initialize()
```
ericbrts, line 237: Instead of calling `ExqaliburLogger.initialize()` everytime, why wouldn't you call a `_initialize()` private function checking what's in the config. Maybe the exqalibur logger isn't even going to need to be initialized, so don't do it in that case.

perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
...
+
+def _my_excepthook(excType, excValue, this_traceback):
+    # only works for the main thread
+    LOGGER.error("Logging an uncaught exception", channel="general",
+                 exc_info=(excType, excValue, this_traceback))
+
+
+def set_level(level: int, channel: str = "user"):
+    CONFIG.set_level(level, channel)
+    LOGGER.set_level()
```
ericbrts, line 249: This line requires parameters!

perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
+# MIT License
```
ericbrts, line 1: This file might need to be separated in at least 2 or 3. Why don't you put all the logger code in a new sub-package (e.g. in perceval/utils/logger)?

perceval/__init__.py
```diff
@@ -66,3 +67,16 @@ def register_plugin(name, silent=False):
     except Exception as e:
         raise RuntimeError("cannot import %s: %s" % (name, str(e)))
     return True
+
+
+def set_logger(use_python_logger: bool = False):
```
ericbrts, line 72: This function shouldn't be here but in the logging package.

And the naming is not readable. Do you expect people to type `pcvl.set_logger(True)` and understand what this means?

I suggest the following naming: pcvl.utils.logging.use_python_logger() and pcvl.utils.logging.use_perceval_logger()

perceval/utils/persistent_data.py
```diff
@@ -170,6 +172,33 @@ def read_file(self, filename: str, file_format: FileFormat) -> Union[bytes, str]
...
+
+    def save_config(self, config: dict):
+        """Save provided token into persistent data, replace any token previously saved
+
+        :param token: token to save
+        """
+        if self.is_writable():
+            file_config = {}
+            if self.has_file(_CONFIG_FILE_NAME):
+                # TODO: add exception for badfileformat
```
ericbrts, line 192: There's a TODO here. Can you do it / remove the comment?

perceval/utils/persistent_data.py
```diff
@@ -170,6 +172,33 @@ def read_file(self, filename: str, file_format: FileFormat) -> Union[bytes, str]
...
+        config = {}
+        if self.has_file(_CONFIG_FILE_NAME):
+            try:
+                config = json.loads(self.read_file(_CONFIG_FILE_NAME, FileFormat.TEXT))
+            except OSError:  # TODO: add exception for badfileformat
+                warnings.warn("Cannot read config file")
+            return config
+
+    def save_config(self, config: dict):
+        """Save provided token into persistent data, replace any token previously saved
```
ericbrts, line 185: Update the docstring `Save the config into ...`

perceval/utils/persistent_data.py
```diff
@@ -170,6 +172,33 @@ def read_file(self, filename: str, file_format: FileFormat) -> Union[bytes, str]
...
+
+        :param token: token to save
+        """
+        if self.is_writable():
+            file_config = {}
+            if self.has_file(_CONFIG_FILE_NAME):
+                # TODO: add exception for badfileformat
+                file_config = json.loads(self.read_file(_CONFIG_FILE_NAME, FileFormat.TEXT))
+            for key in config:
```
ericbrts, line 194: This `for` can be replaced by `file_config.update(config)`

perceval/utils/logging.py
```diff
@@ -0,0 +1,249 @@
...
+                self[_ENABLE_FILE] = config[_ENABLE_FILE]
+
+    def set_level(self, level: int, channel: str = "user"):
+        self[_CHANNELS][channel]["level"] = level
+
+    def save(self):
+        persistent_data = PersistentData()
+        persistent_data.save_config({_LOGGING: dict(self)})
+
+    def message_has_to_be_logged(self, record) -> bool:
```
ericbrts, line 104: Wat?

ericbrts (changes requested):

perceval/utils/persistent_data.py
```diff
@@ -170,11 +180,38 @@ def read_file(self, filename: str, file_format: FileFormat) -> Union[bytes, str]
...
+
+
+
     def clear_all_data(self):
-        """Delete persistent data directory and recreate it
+        """Delete all persistent data
         """
-        shutil.rmtree(self._directory)
-        self._create_directory()
+        for file in os.listdir(self._directory):
```
MarionQuandela, line 213: Don't delete log file

perceval/utils/logging.py
MarionQuandela: I'd like to keep it, it's handy to know where the log are.

PR #404, opened 2024-06-03
Create a QASM to  Perceval `Processor` module

### LINKED ISSUE #399: QASM to linear optics circuit converter
Taking QiskitConverter / MyQLMConverter as examples (located in package `perceval.converter`), write a converter taking a qasm file as input.

* QASM format reference: https://www.quantum-inspire.com/kbase/cqasm/

The QASMConverter will return a fully setup Perceval `Processor`. It needs to support a subset of gates that Perceval already supports (all 1-qubit gates, CX, CZ, CCX and CCZ), qubit naming.
An exception should be raised for other gates, for classical bits, and other feature that Perceval does not support.

maxwell04-wq:


perceval/converters/abstract_converter.py
```diff
@@ -77,6 +77,18 @@ def _configure_processor(self, gate_circuit, **kwargs):
             self._converted_processor.add_port(i * 2, Port(Encoding.DUAL_RAIL, f'{qname}{i}'))
             self._input_list[i * 2] = 1
 
+    def _configure_processor_from_nqubits(self, n_qbits, **kwargs):
+        """
```
raksharuia, line 81: `_configure_processor_from_nqubits` method is identical in action to `_configure_processor` except the required input parameter `n_qubits`. The existing method calls `count_qubits` to get the value. I would suggest using it for the new QASMConverter class as well.

Issue #207, opened 2023-05-24
Heralds out of their box


jsenellart:
**Describe the bug**

Perceval has a renderer at the `Circuit` or the `Processor` level allowing to generate svg or png through matplotlib library. At the processor level, it is possible to add heralds on some modes - as a result, these modes are not anymore available and the heralds are showing inside the circuit.
However, the rendering of the circuit does not take these heralds into account which makes sometime the herald appears as overlapping with the circuit box.

**To Reproduce**
within Pycharm or any editor supporting in-app matplot display, or within a Jupyter notebook
```
import Perceval as pcvl
from perceval.components.core_catalog import PostProcessedCnotItem
pcvl.pdisplay(PostProcessedCnotItem().build(), recursive=True)
```
![image](https://github.com/Quandela/Perceval/assets/96960683/f2b6838f-5eed-4d62-a2f0-8178f3943de8)

or
```
circuit = pcvl.Circuit(3).add(0, pcvl.BS()).add(1, pcvl.BS(1))
processor = pcvl.Processor('SLOS', circuit)
processor.add_herald(0,0)
pcvl.pdisplay(processor, recursive=True)
```
![image](https://github.com/Quandela/Perceval/assets/96960683/2d68308b-83e1-45dd-bfe8-3d3d37ea81ce)

**Expected behavior**
The herald on the first mode should be displayed inside the circuit bounding box

**Additional context**
In notebook, a svg renderer is used (perceval/rendering/canvas/svg_canvas.py), in editor, matplotlib renderer is used (perceval/rendering/canvas/mplot_canvas.py)

alt-shreya:
I'd like to work on this. I have reproduced the error on a Jupyter notebook. What would be a good starting point to go about solving this?

alt-shreya:
I've inspected the code, and I suspect there might be some slight miscalculation in the ```precompute_herald_pos``` function. Checking implementation of ```PERM``` to see what I can find there.

jsenellart:
Hello @alt-shreya, thanks for your interest in the issue 👍 ! I believe that the core issue is coming from the fact the heralds are added *after* the circuit, we would expect them to "push" a bit the top beamsplitter while the circuit is being rendered. 

However how it is done currently:
- the circuit is drawn using `_add_shape` => calling `add_shape` primitives in the canvas renderers
- the heralds are added (so too late to influence the circuit display) using `_update_mode_style`

Let us know if you can figure this out !

(see @raksharuia for further help)

[...]

PR #405, opened 2024-06-05
Fix of heralds display (issue 207)

burlemarxiste:
This is a separate branch for this fix.

https://github.com/Quandela/Perceval/issues/207


ericbrts:
Hello @burlemarxiste ,
I've created an integration branch for UnitaryHACK 2024 contributions. I've just merged your Barrier PR. It seems it created conflicts. Could you please solve them before I review this PR?

I'm assigning the bounty to you right now.

burlemarxiste:
Sure! I will look into this tomorrow!

ericbrts:
Thanks, testing and reviewing it right now!

perceval/rendering/circuit/renderer.py
```diff
@@ -43,12 +43,6 @@ class PortPos:
     def __init__(self, x, y, fixed=True):
```
ericbrts, line 43: I think you can remove the parameter `fixed=True`.

Also, `class PortPos` was meant to follow a herald "Port" moving along it's mode, following permutations. It seems `collect_herald_info` does it for both input and output, and replaces the need for any kind of "changing x and y on the fly". Is that correct?

perceval/rendering/circuit/renderer.py
```diff
@@ -484,114 +575,217 @@ def _add_shape(self, lines, circuit, content, w, shape_fn=None, pos=None):
...
-                p.fixed = False
-        # END Mode tracking + herald positionning algo
-
     def append_circuit(self, lines, circuit, content, pos=None):
         w = self._skin.get_width(circuit)
-        self._add_shape(lines, circuit, content, w, pos=pos)
-        self._update_mode_style(lines, circuit, w)
-        for i in range(lines[0], lines[-1] + 1):
-            self._chart[i] += w
+        if w:
```
ericbrts, line 624: Was this added in order to sometimes use `Barrier` with 0 width?
Anyway, nice add.

perceval/rendering/circuit/renderer.py
```diff
@@ -484,114 +575,217 @@ def _add_shape(self, lines, circuit, content, w, shape_fn=None, pos=None):
...
+                    self._herald_range[1],
+                    self._chart[lines[0] + out_mode] + w)
+            for in_mode, herald_in_mode in input_heralds.items():
+                self._herald_range[0] = min(
+                    self._herald_range[0],
+                    self._chart[lines[0] + in_mode])
+
+    def append_circuit(self, lines, circuit, content, pos=None):
+        w = self._skin.get_width(circuit)
+        self._add_shape(lines, circuit, content, w, pos=pos)
```
ericbrts, line 754: Maybe the `if w:` (from Renderer class) is missing here

perceval/rendering/circuit/renderer.py
```diff
@@ -484,114 +575,217 @@ def _add_shape(self, lines, circuit, content, w, shape_fn=None, pos=None):
...
+        self._update_mode_style(lines, circuit, w)
+        for i in range(lines[0], lines[-1] + 1):
+            self._chart[i] += w
+
+    def append_subcircuit(self, lines, circuit, content):
+        w = self._skin.style_subcircuit['width']
+        self._add_shape(lines, circuit, content, w, self._skin.subcircuit_shape)
```
ericbrts, line 761: Maybe the `if w:` (from Renderer class) is missing here

tests/test_visualization.py
```diff
@@ -322,8 +322,36 @@ def test_svg_processor_with_heralds_phys(tmp_path, save_figs):
...
+
+def test_svg_processor_with_heralds_perm_following_phys(tmp_path, save_figs):
+    c = pcvl.Circuit(4) // (1, PERM([1, 0])) // (1, BS()) // (0, PERM([1, 0])) // BS() // (1, PERM([1, 0]))
+    pc = pcvl.Processor('SLOS', c)
+    pc.add_herald(0,0)
+    pc.add_herald(2,1)
+    _save_or_check(pc, tmp_path, sys._getframe().f_code.co_name, save_figs, recursive=True)
+
+
 def test_svg_processor_with_heralds_and_barriers_phys(tmp_path, save_figs):
```
ericbrts, line 355: This test fails in this branch now that heralds have moved (I think you predicted it in the `Barrier` branch :) )

Also, even when fixing the SVG reference, there seems to be a glitch where the zones left and right of the gray barrier are white (don't take the background color into account)
image

Could you please fix this?

ericbrts (commented):
Thanks for this PR, you made the code better, I'd say!

I'm currently trying to make the pytests run on Github from this PR, hence the target branch change, but I'm kinda failing for now. Anyway, you can run `pytest tests/test_visualization.py` from your computer.

perceval/rendering/circuit/renderer.py
burlemarxiste: Sure, it's not needed, I can remove it.

perceval/rendering/circuit/renderer.py
burlemarxiste: Yes, this was added for my initial implementation that had both "Identity" (acts like a barrier but invisible on the circuit), and "Barrier".

I don't think that's needed anymore, but if you think there might be a use case where some circuit elements have width=0, I can make sure it's used everywhere needed.

tests/test_visualization.py
burlemarxiste: That's strange, the reason why I added `set_background_color` to canvas was precisely to this: made an element aware of the background on which it is drawn. I'll look into it!

burlemarxiste:
I'm a bit lost with the branches at this moment. How should I proceed to continue making changes related to this issue, but with the barrier implementation included to it?

ericbrts:
You fix_heralds branch seems to be the correct source branch to work on. I'm messing with destination branches, because I want to enable the automated test workflow (which I failed to do). In the end, your synchronized with main + your Barrier branch, so working on fix_heralds is the way to go.

I'll merge it in an intermediate branch (right now, named "release/unitary-hack-2024") so that I can control to which scope we add these unitary hack contributions.

burlemarxiste:
Is it okay if I merge the latest version of my barrier branch into this? It would be easier to regenerate the figure for the test that fails. I might also add a few more tests (circuits with `recursive` True or False).

ericbrts:
The Barrier code I merged on Friday is already in your fix_heralds branch (or do you have newer code for Barriers?). I made you split your work into 3 different PR for easier reviews. Now we're reconstructing your work brick by brick. When you merged fix_heralds from the unitary-hack-2024 branch, you retrieved your work about Barriers:
image

burlemarxiste:
Thank you for confirming! I have made the changes you suggested to check for null object `width`, updated the image, and fixed the problem with the missing background color (it was a line that disappeared when I merged with the barrier branch).

burlemarxiste:
I have found an issue with my implementation: the call to `flatten()` in `collect_herald_info` should be done with the same depth of recursion than the diagram is going to be drawn. Can I add an extra `recursion_level` argument to `flatten()` to fix this?

Can you confirm that this is how the diagram should look when `recursive=False`?

Screenshot 2024-06-10 at 17 13 01

ericbrts:
Yes you can add a `recursion_level` parameter with a default value meaning max recursion level (like 0 ? -1 ?)

The picture for the non-recursive display is totally fine.

burlemarxiste:
Thanks for confirming that the image is correct.

I just understood something that answered my initial confusion about the `recursive` argument: **there are only 2 possible recursion levels.**  The reason is that when a Processor is added to a Processor, they are composed (no nesting between them); and iterating on a circuit always goes to the deepest level.

So there are only two possible levels, and there is no needed for an extra argument. I just need to collect the herald info at the component level when `recursive=False`, and circuit element level when `recursive=True`.

This latest commit should fix it.

ericbrts:
Thanks @burlemarxiste , I'm reviewing / testing again now. Do you wish to add other things to your PR, or can I merge once I approve it?

ericbrts:
Yes please, file as many issues as you wish!

PR #408, opened 2024-06-05
PCVL-725 ResourcesEstimator

JuanJoseOchoaD:
circuit_to_graph_converter is used to convert a circuit into a weighted graph, whose gates will have a certain weight corresponding to the actual non-deterministic gates that would have to be performed in the optical circuit. It also has a plotting function and an algorithm to cut it in k subgraphs in the most optimal way.

The second file, resources_estimator calculates for a given qudit encoding, including the ones suggested by circuit_to_graph_converter, the number of entangling gates, photons and modes needed.


perceval/converters/circuit_to_graph_converter.py
```diff
@@ -0,0 +1,188 @@
...
+    gates_qubits = []
+    for instr in qiskit_circuit.data:
+        # Extract gate name and qubits
+        gates_names.append(instr[0].name)
+        qubits = instr[1]
+        qubit_numbers = [qiskit_circuit.find_bit(q).index for q in qubits]
+        gates_qubits.append(qubit_numbers)
+    return gates_names, gates_qubits
+
+
```
raksharuia, line 59: Suggestion: include `CircuitToGraphConverter` class in the `__init__` of converters following other converter classes. It will allow you import by using `from perceval.converters import CircuitToGraphConverter`

perceval/utils/qmath.py
```diff
@@ -61,3 +64,48 @@ def exponentiation_by_squaring(base, power: int):
         temp_base = temp_base * temp_base
 
     return result
+
+
+def kmeans(features: np.ndarray, n_clusters: int, n_init: int = 10) -> np.ndarray:
+    """
+    Manual KMeans implementation.
```
raksharuia, line 71: Please add a little description of what `KMeans` is and/or some reference.

perceval/utils/qmath.py
```diff
@@ -61,3 +64,48 @@ def exponentiation_by_squaring(base, power: int):
...
+
+    :param features: Data points for clustering.
```
raksharuia, line 73: parameter type is already written in the function declaration (L69). Remove the `:type` from docstring. Could you please do this for all other methods too wherever `:type` is written in docstrings.

perceval/utils/qmath.py
```diff
@@ -61,3 +64,48 @@ def exponentiation_by_squaring(base, power: int):
...
+    """
+    best_labels = None
+    best_inertia = float('inf')
+
+    for _ in range(n_init):
+        # Initialize centroids randomly
+        indices = np.random.choice(features.shape[0], n_clusters, replace=False)
+        centroids = features[indices]
+
+        for _ in range(300):  # Maximum number of iterations
```
raksharuia, line 90: This kind of hard-coded values can cause bugs. The usual way we implement them is by `VARIABLE_NAME=300`, then use `VARIABLE_NAME` in place of the hard coded value

raksharuia (changes requested):

tests/test_circuit_to_graph_resources_estimator.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # Generate the graph
+    graph = converter.graph_generator()
+
+    # Plot the graph (optional)
+    # converter.plot_graph(graph)
+
+    # Check calculation
+    single_calc_result = converter.graph_k_clustering_and_cnots_needed()[0]
+    cx_count = converter.graph_k_clustering_and_cnots_needed()[1]
```
raksharuia, line 69: L68 and L69 would call the method twice and do all the computation twice. You can extract both variables in one line by using `single_calc_result, cx_count = converter.graph_k_clustering_and_cnots_needed()`.

perceval/converters/circuit_to_graph_converter.py
```diff
@@ -0,0 +1,188 @@
...
+        possible_partitions = []
+        number_of_cnots = []
+
+        for k in range(2, n + 1):
+            # 1 is the trivial case of the complete circuit in one qudit and n is the dual rail.
+            features = eigenvectors[:, :k]
+            if compute_with_min_cnots:
+                best_partition = None
+                min_cnots = float('inf')
+                for _ in range(30):
```
raksharuia, line 152: This hard-coded value of `30` should be replaced by variable name to increase readability of the code. If it is some fixed number, I suggest adding `CAPTILIZED_VARIABLE_NAME=30` at the beginning of the file (outside class scope) and then using it here.

raksharuia (changes requested):

perceval/converters/circuit_to_graph_converter.py
JuanJoseOchoaD: Is it normal that when I try to run the __init__ after including it, I get 
Traceback (most recent call last):
  File "C:\DEV\Perceval\perceval\converters\__init__.py", line 30, in <module>
    from .qiskit_converter import QiskitConverter
ImportError: attempted relative import with no known parent package
?

perceval/rendering/pdisplay.py
```diff
@@ -327,6 +327,22 @@ def pdisplay_tomography_chi(qpt: ProcessTomography, output_format: Format = Form
     plt.show()
 
 
+def pdisplay_graph(g: nx.Graph, output_format: Format = Format.MPLOT):
+    if output_format not in {Format.MPLOT, Format.LATEX}:
+        raise TypeError(f"Graph plot does not support {output_format}")
+    if output_format == Format.LATEX:
+        print(nx.to_latex(g))
```
ericbrts, line 334: I think this should be `return nx.to_latex(g)`.
The higher layers of pdisplay will do the display.

perceval/rendering/pdisplay.py
```diff
@@ -327,6 +327,22 @@ def pdisplay_tomography_chi(qpt: ProcessTomography, output_format: Format = Form
     plt.show()
 
 
+def pdisplay_graph(g: nx.Graph, output_format: Format = Format.MPLOT):
+    if output_format not in {Format.MPLOT, Format.LATEX}:
```
ericbrts, line 331: What about svg format?

perceval/utils/qmath.py
```diff
@@ -61,3 +64,45 @@ def exponentiation_by_squaring(base, power: int):
...
+    Manual KMeans implementation. Clusterizes the system in k subsets.
+
+    :param features: Data points for clustering.
+    :param n_clusters: Number of clusters.
+    :param n_init: Number of times the k-means algorithm will be run with different centroid seeds.
+    :return: Cluster labels.
+    :rtype: np.ndarray
+    """
+    best_labels = None
+    best_inertia = float('inf')
```
ericbrts, line 80: Isn't it `np.inf`?

perceval/converters/circuit_to_graph_converter.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        if gates is not None and qubits is not None:
+            self.gates = gates
+            self.qubits = qubits
+        elif qiskit_circuit is not None:
+            self.gates, self.qubits = gates_and_qubits(qiskit_circuit)
+        else:
+            raise ValueError("Either a Qiskit circuit or both gates and qubits lists must be provided")
+
+    def graph_generator(self) -> nx.Graph:
```
ericbrts, line 82: This method should be names `generate_graph` (verb) instead of a substantive.

perceval/converters/circuit_to_graph_converter.py
```diff
@@ -0,0 +1,187 @@
...
+                            edge_weights[edge] += 2
+                        else:
+                            edge_weights[edge] = 2
+
+        for edge, weight in edge_weights.items():
+            g.add_edge(*edge, weight=weight)
+        return g
+
+    @staticmethod
+    def plot_graph(g):  # g is the graph created in graph_generator
```
ericbrts, line 115: Shouldn't you remove this function now that pdisplay is ready to take a graph as parameter.

perceval/converters/resources_estimator.py
```diff
@@ -0,0 +1,90 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from perceval.converters.circuit_to_graph_converter import CircuitToGraphConverter, gates_and_qubits
+import numpy as np
+from qiskit import QuantumCircuit
```
ericbrts, line 32: This imposes `qiskit` as a primary 3rd party for Perceval. However, it's just an "extra" requirement. So, the import shouldn't be done at the top level of this class.

What we need: `import perceval` should keep working, without `qiskit` being installed on the venv.

ericbrts (changes requested):

perceval/converters/resources_estimator.py
JuanJoseOchoaD: Indeed I don't need qiskit here, I just have it to define the circuit as a qiskit_circuit type. Then, can I just remove it and write something like 

 def __init__(self, qiskit_circuit, encoding: list[list[int]] = None):
        """
       :param qiskit_circuit: Qiskit quantum circuit to estimate resources for.
       :param encoding: Custom encoding to use.
        """ 
 ?

tests/test_circuit_to_graph_resources_estimator.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_circuit_to_graph_converter():
+    qiskit_circuit = random_circuit(8, 10, max_operands=2)
+    converter = CircuitToGraphConverter(qiskit_circuit=qiskit_circuit)
+
+    # Generate the graph
+    graph = converter.graph_generator()
+
+    # Plot the graph (optional)
+    # converter.plot_graph(graph)
```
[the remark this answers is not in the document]
raksharuia: Delete the lines that are commented out here. They are not needed in the test.

raksharuia (approved):

perceval/rendering/pdisplay.py
JuanJoseOchoaD: SVG is not supported by networkx

ericbrts (approved):

PR #419, opened 2024-06-25
Pcvl 742 entice users to use min detected photon filter

MarionQuandela:


ericbrts (approved):

ericbrts:
I approved, however, could you please write a few lines in legacy.rst to explain users that not setting a filter value for noisy simu or on the Cloud is now a deprecated behavior?

PR #430, opened 2024-07-12
change rpc handler code and add tests

Quandevla:
A little rewrite of the RPC Handler with some tests - using requests-mock !

Remember to install `requests-mock`

```sh
python -m pip install -r tests/requirements.txt
# or
python -m pip install requests-mock
```

Useful commands to run tests
```sh
# for all tests
python -m pytest --cov-report html:dist --cov=perceval

# for one file - here tests/test_payload_generation.py
python -m pytest --cov-report html:dist --cov=perceval tests/test_payload_generation.py
```

To run precommit
```sh
python -m pre_commit run --all-files
```


perceval/runtime/rpc_handler.py
```diff
@@ -39,57 +41,94 @@
...
-        return f"{self.url}{endpoint}"
+    def build_endpoint(self, endpoint, *args):
+        """build the default endpoint url
+
+        :param endpoint: first part of the endpoint
+        :return: the full endpoint url from the args
+        """
+        endpath = ''
+        if len(args) > 0:
+            endpath = f"/{'/'.join(str(x) for x in args)}"
```
ericbrts, line 71: `endpath` always starts with a slash. When endpoint ends with a slash too, there's a `//` in the resulting address.
If you look at the `_ENDPOINT_<whatever>` constants, they all finish with a slash.
So I guess that's a problem.

perceval/runtime/rpc_handler.py
```diff
@@ -39,57 +41,94 @@
...
+        :param token: token used for identification
+        """
         self.name = name
         self.url = url
         self.token = token
-        self.headers = {
-            'Authorization': f'Bearer {token}'
-        }
+        self.headers = {'Authorization': f'Bearer {token}'}
+        self.def_t = 10  # default timeout
```
ericbrts, line 61: This variable name shouldn't be named `def_t`. Indeed, 10 is the default value for the request timeout, but the variable itself doesn't embody a default, it can be changed a the `RPCHandler` user.

I suggest you rename this variable `timeout` or `request_timeout`

Quandevla: modified in my lastest commit https://github.com/Quandela/Perceval/pull/430/commits/e3816c9200156966a2d0454feb81e04ea5e37f6a

PR #431, opened 2024-07-12
Pcvl 750 improved random computation test for photon recycling

raksharuia:


tests/test_loss_mitigation.py
```diff
@@ -70,3 +75,64 @@ def test_photon_loss_mitigation(result_type):
     for keys, value in mitigated_dist.items():
         assert sum(keys) == ideal_photon_count
         assert value > 1e-6
+
+def _compute_random_circ_probs(source_emission, num_photons):
+
+    mzi = (BS() // (0, PS(phi=Parameter("φ_a")))
+           // BS() // (1, PS(phi=Parameter("φ_b"))))
+
+    random_loc = Circuit.decomposition(Matrix.random_unitary(20), mzi,
```
ericbrts, line 84: Why do you need to decompose your unitary matrix into an actual interferometer?
I think you just need:
```python
random_loc = Unitary(Matrix.random_unitary(20))
``` 
It'll run a lot faster and simulate the same thing (minus the floating point inaccuracy injected by the `decomposition`).

tests/test_loss_mitigation.py
```diff
@@ -70,3 +75,64 @@ def test_photon_loss_mitigation(result_type):
...
+                                              phase_shifter_fn=PS,
+                                              shape="triangle")
+    # Processor config
+    processor = Processor("SLOS", random_loc)
+    processor.min_detected_photons_filter(0)
+    processor.thresholded_output(True)
+
+    # Source properties
+    src = Source(emission_probability=source_emission)
+    processor.source = src
```
ericbrts, line 94: Please use `NoiseModel` instead of `Source` as it's deprecated.

A way of doing this is when you instantiate the Processor:
```python
processor = Processor("SLOS", random_loc, noise=NoiseModel(transmittance=source_emission))
```

tests/test_loss_mitigation.py
```diff
@@ -70,3 +75,64 @@ def test_photon_loss_mitigation(result_type):
...
+
+    for key, values in ideal_dist.items():
+        if key not in mitigated_dist.keys():
+            mitigated_dist[key] = 0.0
+
+    # computing tvd
+    tvd_miti = total_variation_distance(list(ideal_dist.values()), list(mitigated_dist.values()))
+    tvd_ps = total_variation_distance(list(ideal_dist.values()), list(post_dist))
+
+    assert tvd_miti , tvd_ps  # checks that mitigated is closer to ideal than post-selected distribution
```
ericbrts, line 138: What does this assert do?
Doesn't it check that the tuple `tvd_miti , tvd_ps` (yes it's a tuple) is not empty. Which it cannot be, as it has 2 elements by construction.

raksharuia: This assert is my typo, I entered ', ' instead of '<' check. A good catch :-)

tests/test_loss_mitigation.py
raksharuia: Yes, you are correct. I did not need it. I had it from the notebook in my scripts and missed to remove it. Fixed.

tests/test_loss_mitigation.py
raksharuia: Sorry, my bad. I forgot. I took this and updated the existing test to use `NoiseModel()` too.

PR #437, opened 2024-07-17
PCVL-739-improved checks at entry of recycling

raksharuia:


tests/test_loss_mitigation.py
```diff
@@ -70,3 +70,38 @@ def test_photon_loss_mitigation(result_type):
...
+    # missing n-2 state
+    lossy_bsd2 = BSDistribution()
+    lossy_bsd2[bs1] = 0.3333
+    lossy_bsd2[bs2] = 0.3333
+    lossy_bsd2[bs4] = 0.3333
+
+    with pytest.raises(ValueError):
+        photon_recycling(perfect_bsd, 3)
+
+        photon_recycling(lossy_bsd1, 3)
```
ericbrts, line 103: Each failing statement should have its own `with pytest.raises(ValueError):`, else it's like saying "I want at least one statement to raise a ValueError" when you actually expect **all** statements to raise.

perceval/error_mitigation/loss_mitigation.py
```diff
@@ -36,6 +36,22 @@
 from ._loss_mitigation_utils import _gen_lossy_dists, _get_avg_exp_from_uni_dist, _generate_one_photon_per_mode_mapping
 
 
+def _validate_noisy_input(noisy_input: Union[BSCount, BSDistribution], ideal_photon_count: int):
```
ericbrts, line 39: Should we await the theoretician point of view on this function content?

PR #438, opened 2024-07-22
Pcvl 751 perceval wrapper for rectangular decomposer

MarionQuandela:


perceval/utils/algorithms/rectangular_decomposer.py
```diff
@@ -0,0 +1,76 @@
...
+from perceval.serialization import deserialize_circuit
+
+
+class RectangularDecomposer:
+    """Instantiate a rectangular Decomposition
+
+        :param precision: precision of the decomposition, defaults to 1e-6
+        """
+
+    def __init__(self, precision: float = 1e-6):
```
Benoit-F-Q, line 44: I'm not found of default values for parameters, espacially when there are 3 nested levels of default value (RectangularDecomposerWrapper and RectangularDecomposer have their own default values)

Benoit-F-Q: Moreover, 1e-6 is a very low value. All tests should pass with 1e-16

perceval/utils/algorithms/rectangular_decomposer.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_interferometer(self, add_phase_correction: bool) -> ACircuit:
+        """get the rectangular interferometer
+
+        :param add_phase_correction: whether to add the final vertical line of phase shifters to the final interferometer.
+
+        :return: the rectangular interferometer
+        """
+        return deserialize_circuit(self.rectangular_decomposer.get_interferometer(add_phase_correction))
+
+    def set_precision(self, precision: float) -> None:
```
Benoit-F-Q, line 72: Setter+getter, une property ferai l'affaire, non ?

PR #440, opened 2024-07-24
added Session constructors **kwargs for better compatibility

oradomskyi:
Without **kwargs user gets somewhat locked on required named arguments of a particular provider, that requires a user to take care about handling both _required_ **and** _unexpected_ arguments.

In particular, in PennyLane-Quandela plugin, a Device class takes **kwargs for having a cleaner user interface and a logic that creates backend and then, if provider name exists in kwargs, it instantiates a session with a cloud provider.

Without **kwargs added to ISession either would happen - an interface of the Device would get bigger for taking into account specifics of provider creation or an ISession will fail with "_unexpected argument_" or "_multiple values for the same argument_"

https://github.com/oradomskyi/pennylane-quandela/blob/main/pennylane_quandela/device.py#:~:text=def%20__init__

### DIFF
--- a/perceval/providers/quandela/quandela_session.py
+++ b/perceval/providers/quandela/quandela_session.py
@@ -33,7 +33,7 @@ from perceval.runtime.rpc_handler import RPCHandler
 
 
 class Session(ISession):
-    def __init__(self, platform_name: str, token: str, url: str = None, **kwargs):
+    def __init__(self, platform_name: str, token: str, url: str = None):
         self._platform_name = platform_name
         self._token = token
         self._url = url or QUANDELA_CLOUD_URL
--- a/perceval/providers/quandela/quandela_session.py
+++ b/perceval/providers/quandela/quandela_session.py
@@ -33,7 +33,7 @@ from perceval.runtime.rpc_handler import RPCHandler
 
 
 class Session(ISession):
-    def __init__(self, platform_name: str, token: str, url: str = None):
+    def __init__(self, platform_name: str, token: str, url: str = None, **kwargs):
         self._platform_name = platform_name
         self._token = token
         self._url = url or QUANDELA_CLOUD_URL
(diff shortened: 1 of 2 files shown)


oradomskyi:
A failing test of loss mitigation function is unrelated to the proposed changes.

ericbrts:
Hi @oradomskyi ,

I think it's bad practice to add `**kwargs` in sibling classes, just to be sure you can instantiate one with the parameters from the other. It'll make the code less maintainable.

I've checked your code from https://github.com/oradomskyi/pennylane-quandela/blob/main/pennylane_quandela/device.py and it seems you're trying to build the `Session` object from within your class. Would it be possible to ask the user to create the `Session` before creating the `QuandelaDevice`?
See : https://en.wikipedia.org/wiki/Dependency_injection

You'd get something along the lines of:
```python

session = pcvl.providers.any_provider.Session(specific_param1, specific_param2, ..., specific_paramN)
pennylane_quandela.QuandelaDevice(..., session, ...)  # Here you won't need **kwargs to construct the session as it's already been taken care of
```

Tell me what you think.

Thanks,
Eric

oradomskyi:
Hi @ericbrts , this is done - now user would bear responsibility for establishing a session with the cloud :)

My decision to hide Session away from user was based on experience of making libs for users who are not that versed in programming and who need as simple and standardized interface as possible.

Let me know if feeding the session_id is ok for you, it is for cases when a user created session using the cloud dashboard. 

Best,
Sasha

ericbrts:
Thanks @oradomskyi ,

I'll have to check with Scaleway on your remaining change, because I'm not aware of this session ID retrieval.

oradomskyi:
I looked at https://www.scaleway.com/en/developers/api/qaas/ and ScalewayProvider seem to implement the session creation when calling start()

Issue seem to be on the side of my Scaleway account, so wrote to the Support.

oradomskyi:
I propose to discard this pull request and changes, and I will create a fresh one with updates for ScalewayProvider if I will figure something out.

What do you think about using the Scaleway SDK for qaas api calls instead of in-house implementation?

ericbrts:
Hi @oradomskyi ,

Ok to close this pull request for now.

The Perceval implement of the Scaleway provider classes come directly from Scaleway people. So, I need to contact them in order to get their opinion.

PR #443, opened 2024-07-30
Pcvl 766 user channel logs

MarionQuandela:


MarionQuandela:
- Replace all UserWarnings by logger.warn
- Implement a deprecated decorator and use it instead of deprecated.deprecated
- Adapt UT by patching log.warn and use a similar syntax as before:
`with WarnLogChecker(mock_warn):`
instead of
`with pytest.warns(UserWarning):`

tests/test_utils_conversion.py
```diff
@@ -68,13 +75,13 @@ def test_sample_count_to_probs():
     })
     output = sample_count_to_probs(sample_count)
     assert sum(output.values()) == 1
-    assert output[b0] == 0.28
-    assert output[b1] == 0.12
-    assert output[b2] == 0.4
-    assert output[b3] == 0.2
+    assert output[b0] == pytest.approx(0.28)
```
raksharuia, line 78: 💯

tests/_test_utils.py
```diff
@@ -193,29 +195,23 @@ def _save_or_check(c, tmp_path, circuit_name, save_figs, recursive=False, compac
...
-        assert_sv_close(sv1, sv4)
-    except AssertionError:
-        print("detected sv are different")
-
-    try:
-        assert_sv_close(sv4, sv1)
-    except AssertionError:
-        print("detected sv are different")
+class WarnLogChecker():
+    def __init__(self, mock_warn: MagicMock, log_channel: channel = channel.user, expected_log_number: int = 1):
```
raksharuia, line 199: 👍

tests/_test_utils.py
```diff
@@ -193,29 +195,23 @@ def _save_or_check(c, tmp_path, circuit_name, save_figs, recursive=False, compac
         assert ok, msg
 
 
-if __name__ == "__main__":
```
raksharuia, line 196: The tests under ` main` here are testing some of the functions written inside `_test_utils`. Could you create a new test file `test_test_utils.py` and move them there ?

raksharuia (changes requested):

perceval/components/abstract_processor.py
```diff
@@ -560,10 +561,10 @@ def check_input(self, input_state: BasicState):
...
 
     def _deduce_min_detected_photons(self, expected_photons: int) -> None:
-        warnings.warn(UserWarning(
-            f"Setting a value for min_detected_photons will soon be mandatory, please change your scripts accordingly\n" +
-            "Use the method processor.min_detected_photons_filter(value) before any call of processor.with_input(input)\n" +
-            "The current deduced value of min_detected_photons is {expected_photons}"))
+        logger.warn(
+            "Setting a value for min_detected_photons will soon be mandatory, please change your scripts accordingly." +
+            " Use the method processor.min_detected_photons_filter(value) before any call of processor.with_input(input)." +
+            f" The current deduced value of min_detected_photons is {expected_photons}", channel.user)
```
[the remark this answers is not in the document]
MarionQuandela: We don't want any warning in case of perfect source (no noise) in local simulation, so it's normal. As for the two warnings with local processor and an imperfect source and not setting min_detected_photons_filter, it's also the expected behavior :)

ericbrts (approved):

Issue #414, opened 2024-06-11
Rendering: incorrect size of circuit box


burlemarxiste:
When a processor contains nested circuits (created with the `merge=False` argument passed to `add`), the blue box enclosing each subblock is of the wrong size.

The reason is that the drawing of elements stops at one level of recursion, whereas the `get_size` call goes to the full depth.

This is a circuit from a tutorial:

```python
import perceval as pcvl
from perceval.components import core_catalog
from perceval.rendering import format

from perceval.components.unitary_components import PS, BS, PERM

pre_MZI = (pcvl.Circuit(4, name="Bell State Prep")
           .add(0, BS())
           .add(2, BS())
           .add(1, PERM([1, 0])))

upper_MZI = (pcvl.Circuit(2, name="upper MZI")
             .add(0, PS(phi=pcvl.P('phi_0')))
             .add(0, BS())
             .add(0, PS(phi=pcvl.P('phi_2')))
             .add(0, BS()))

lower_MZI = (pcvl.Circuit(2, name="lower MZI")
             .add(0, PS(phi=pcvl.P('phi_1')))
             .add(0, BS())
             .add(0, PS(phi=pcvl.P('phi_3')))
             .add(0, BS()))

chip = (pcvl.Circuit(4)
              .add(0, pre_MZI, merge=False)
              .add(0, upper_MZI, merge=False)
              .add(2, lower_MZI, merge=False))

processor = pcvl.Processor('SLOS', chip)
output_format = format.Format.MPLOT
extension = "pdf"

pcvl.pdisplay_to_file(processor, 'pdisplay_multiblocks.' + extension, output_format, recursive=True)
```

By changing some of the `merge=False` arguments to `True`, one can get circuits with various degrees of nesting. In all the following examples, the size of the blue circuit box is the same, and corresponds to the size of the fully flattened circuit.

<img width="1295" alt="Screenshot 2024-06-11 at 10 49 48" src="https://github.com/Quandela/Perceval/assets/171129116/b9b563b9-28ef-4801-846b-55b806715b32">
<img width="1263" alt="Screenshot 2024-06-11 at 10 50 27" src="https://github.com/Quandela/Perceval/assets/171129116/e36a8dd0-8a1e-46cc-94e8-feacf7e6fc81">
<img width="637" alt="Screenshot 2024-06-11 at 10 50 47" src="https://github.com/Quandela/Perceval/assets/171129116/d7aa3e17-4a1d-4251-a038-522a42d2c203">

[...]

PR #444, opened 2024-07-30
fix for issue Rendering: incorrect size of circuit box #414

oradomskyi:
Setting recursive=False in the inner loop of circuit rendering seems to fix the issue, txt pdf and html files are generated correctly and contain the correct circuit of proper size.

image

files generated for the circuit mentioned in the original issue are here:
https://drive.google.com/drive/folders/1E1cF_RRrfD8HEcfvpSkX1at810IYkHhl?usp=sharing

### DIFF
--- a/perceval/rendering/circuit/renderer.py
+++ b/perceval/rendering/circuit/renderer.py
@@ -104,7 +104,7 @@ class ICircuitRenderer(ABC):
                     if c.is_composite() and c._components:
                         if recursive:
                             self._current_subblock_info = self._subblock_info.setdefault(c, {})
-                            self.open_subblock(shiftr, c.name, self.get_circuit_size(c, recursive=True), c._color)
+                            self.open_subblock(shiftr, c.name, self.get_circuit_size(c, recursive=False), c._color)
                             self.render_circuit(
                                 c,
                                 shift=shiftr[0],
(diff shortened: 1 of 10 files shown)


oradomskyi:
image

image

oradomskyi:
PR is related to the box size, not the positioning of heralds - that is another issue

raksharuia:
Yes, fixing of heralds (changes have already been made for this) is another issue; your fix corrects the issue of incorrect box size on rendering. Thanks for the contribution :-)

perceval/rendering/circuit/renderer.py
```diff
@@ -104,7 +104,7 @@ def render_circuit(self,
                     if c.is_composite() and c._components:
                         if recursive:
                             self._current_subblock_info = self._subblock_info.setdefault(c, {})
-                            self.open_subblock(shiftr, c.name, self.get_circuit_size(c, recursive=True), c._color)
+                            self.open_subblock(shiftr, c.name, self.get_circuit_size(c, recursive=False), c._color)
                             self.render_circuit(
```
raksharuia, line 108: Could you please add a unittest (following our conventions on testing circuit images) using the example from the issue?

raksharuia (changes requested):

perceval/rendering/circuit/renderer.py
oradomskyi: This is done, let me know if you would like to see other circuit tests.

P.S. This issue seems to be specific to Canvas Renderer. Text Renderer appears to have no problems with box size.

oradomskyi: Failing tests on Windows does not seem to be related to the proposed changes.

PR #448, opened 2024-08-06
PCVL-768 general channel logs

ericbrts:


MarionQuandela (commented):
I would also add some other log:
- In utils/algorithms/simplification.py as debug level (@raksharuia could do it in her PR)
- In simulators/noisy_sampling_simulator.py ?
- Add warnings regarding Polarization ?

perceval/utils/_random.py
```diff
@@ -41,3 +43,9 @@ def random_seed(seed=None):
     """
     random.seed(seed)
     np.random.seed(seed)
+    if seed is None:
+        logger.info(f"Reset RNG to a time-based random seed", channel.general)
```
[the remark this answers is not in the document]
ericbrts: I guess we don't, because in this case, it's a different seed value in `numpy`, pure python `random` and `exqalibur`

perceval/components/abstract_processor.py
```diff
@@ -35,9 +35,8 @@
```
raksharuia, line 35: I think it will be useful to add some `debug` level logs for the method `_compose_processor()` (L263). Could you some there as well?

Sorry Github is not letting me put inline comments there.

raksharuia (changes requested):

perceval/components/abstract_processor.py
```diff
@@ -261,14 +261,17 @@ def _validate_postselect_composition(self, mode_mapping: Dict):
                 f"Post-selection conditions cannot compose with modes {impacted_modes}"
 
     def _compose_processor(self, connector: ModeConnector, processor, keep_port: bool):
+        logger.debug(f"Compose processor {self.name} with {processor.name}", channel.general)
         self._is_unitary = self._is_unitary and processor._is_unitary
         self._has_td = self._has_td or processor._has_td
         if processor.heralds and not processor.parameters:
             # adding the same processor component again renders incorrect heralds if not copied
             # This concerns our gate based processors from catalog which has no input params
+            logger.debug("  Force copy during processor compose", channel.general)
```
raksharuia, line 270: Extra space at the beginning of the message. It seems intentional for readability - something like an indent for subsequent debug messages inside. I am wondering if there is any other way to implement this instead of typing the 'space' ?

oradomskyi: I don't think so, some function will have to eventually add these spaces or tab symbols.

ericbrts: We could implement an "indent" parameter (as a positive integer) but I'd say it's at least another task, and I'm not so sure we'd need it.

PR #451, opened 2024-08-14
PCVL-780 Barrier can be invisible

ericbrts:


perceval/components/unitary_components.py
```diff
@@ -453,30 +441,34 @@ def __init__(self):
...
-        super().__init__(U=u)
+    def __init__(self, num_modes: int, visible: bool = True):
+        assert isinstance(num_modes, int), "Barrier"
+        self._visible = bool(visible)
+        super().__init__(num_modes)
 
-    def describe(self, _=None):
-        return f"Barrier({self.num_modes})"
+    @property
+    def visible(self):
```
MarionQuandela, line 459: Maybe add a setter also to allow the user to have invisible barriers ?

ericbrts: `visible` is already a parameter of `Barrier.__init__`

MarionQuandela: Yes but when you create a barrier with the operator @ you don't have the possibility to make it invisible

perceval/components/unitary_components.py
```diff
@@ -453,30 +441,34 @@ def __init__(self):
...
+    def _compute_unitary(self, assign: dict = None, use_symbolic: bool = False) -> Matrix:
+        return Matrix.eye(self._m)
+
+    def describe(self):
+        return f"Barrier({self._m})"
 
     def definition(self):
         return self.U
 
+    # noinspection PyMethodMayBeStatic
```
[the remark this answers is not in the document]
ericbrts: En effet, j'avais pas remarqué. Mais ce ne sont pas les premiers "noinspection" du code

tests/test_generic_interferometer.py
```diff
@@ -40,6 +40,19 @@
 mzi_generator_func = catalog['mzi phase last'].generate
 
 
+# noinspection PyTypeChecker
```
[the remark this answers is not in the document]
ericbrts: I just moved this code. This "noinspection" mark was moved with it. I've just deleted it and it adds no warning. It seems it was useless.

perceval/components/unitary_components.py
ericbrts: You don't have to use @ operator. It's only a facilitator for a very very specific case

PR #452, opened 2024-08-23
Pcvl 769 logging user documentation

MarionQuandela:


perceval/utils/logging/loggers.py
```diff
@@ -123,6 +126,15 @@ def _configure_logger(self):
                     exq_log.level.__members__[level],
                     exq_log.channel.__members__[channel])
 
+    def apply_config(self, config: LoggerConfig):
+        if config.python_logger_is_enabled():
+            self = PythonLogger()
```
ericbrts, line 131: Could you test this thoroughly, please?

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,197 @@
...
+     - ``logger.debug``
+     - Detailed information, typically of interest only when diagnosing problems.
+   * - INFO
+     - ``logger.info``
+     - Confirmation that things are working as expected.
+   * - WARNING
+     - ``logger.warn``
+     - An indication that something unexpected happened, or indicative of some problem in the near future (e.g. ‘disk space low’). The software is still working as expected.
+   * - ERROR
+     - ``logger.error``
```
ericbrts, line 85: It seems we called it "err" in exqalibur and not error :/

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,197 @@
...
+   :header-rows: 1
+   :stub-columns: 1
+   :width: 100%
+   :align: center
+
+   * - Log level
+     - Perceval call
+     - Usage
+   * - DEBUG
+     - ``logger.debug``
```
ericbrts, line 76: Why \`\`logger.debug\`\` and not \`\`level.debug\`\` ?

docs/source/reference/logging.rst
MarionQuandela: Yes but the python logger module call it error. I think it's a good idea to keep the same interface :)

perceval/utils/logging/config.py
```diff
@@ -50,6 +54,11 @@ def _init_channel(self, channel: exqalibur_logging.channel, level: exqalibur_log
         self[_CHANNELS][channel.name]["level"] = level.name
 
     def reset(self):
+        """Reset the logger configuration to it's default value meaning:
```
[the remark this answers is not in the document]
MarionQuandela: I replaced meaning by "which is"

docs/source/reference/logging.rst
MarionQuandela: Yes but the python logger module call it error. I think it's a good idea to keep the same interface :)

docs/source/reference/logging.rst
MarionQuandela: Yes but the python logger module call it error. I think it's a good idea to keep the same interface :)

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,200 @@
+Logging
+=======
+
+To log with Perceval you can either a built-in Perceval logger or the python one. By default, our logger will be used.
```
raksharuia, line 4: ... you can either 'use' a built-in ... The word 'use' is missing,

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,200 @@
...
+
+To have the perceval message log with the python logger, use this method:
```
raksharuia, line 6: To log the Perceval message logs with the python logger, use this method:

(the word log gets too repetitive I understand but 'have' seemed weird to me, maybe 'save'?)

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,200 @@
...
+
+.. code-block:: python
+
+    from perceval.utils import use_python_logger
+    use_python_logger()
+
+To use back the perceval logger use this method:
```
raksharuia, line 13: To switch back to using perceval logger, use:

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,200 @@
...
+
+   * - Channel
+     - Default level
+     - Usage
+   * - ``general``
+     - off
+     - General info: Technical info, track the usage of features
+   * - ``resources``
+     - off
+     - Info about how are use our backends or remote platform GPU use (exqalibur)
```
raksharuia, line 124: Usage info about our backends or remote platform GPU (exqalibur)

docs/source/reference/logging.rst
```diff
@@ -0,0 +1,200 @@
...
+    from perceval.utils import logger, logging
+    logger_config = logging.LoggerConfig()
+    logger_config.enable_file()
+    logger_config.set_level(logging.level.info, logging.channel.user)
+    logger.apply_config(logger_config)
+
+Log format
+----------
+
+In the console the log will appear with the format:
```
raksharuia, line 190: **On** the console...

raksharuia (changes requested):

docs/source/reference/logging.rst
ericbrts: What I meant is: we use `exqalibur.logger.level` enum in Python, and exqalibur exposes the error level as `exqalibur.logger.level.err` which is exposed by Python as `perceval.utils.logging.level.err`

So isn't there an issue with the doc?

perceval/utils/logging/loggers.py
```diff
@@ -123,6 +126,13 @@ def _configure_logger(self):
                     exq_log.level.__members__[level],
                     exq_log.channel.__members__[channel])
 
+    def apply_config(self, config: LoggerConfig):
```
ericbrts, line 129: To avoid the inability to change the logger type, from inside a given logger, you can define an `apply_logger_config()` function outside of the logger classes. It would remove both user warnings.

docs/source/reference/logging.rst
ericbrts: To log a message with the Python logger, use:

PR #455, opened 2024-08-29
Pcvl 771 deprecated min detected photon filter

MarionQuandela:


perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -159,13 +165,23 @@ def set_circuit(self, circuit: ACircuit):
...
     def set_min_detected_photon_filter(self, value: int):
         """
         Set the physical detection filter. Any output state with less than this threshold gets discarded.
 
         :param value: Minimal photon count in output states of interest.
         """
-        self._min_detected_photon_filter = value
+        self._min_detected_photons_filter = value
+
+    def set_min_detected_photons_filter(self, value: int):
```
raksharuia, line 178: 💯

perceval/simulators/simulator.py
```diff
@@ -78,13 +78,23 @@ def precision(self, value: float):
...
     def set_min_detected_photon_filter(self, value: int):
         """
         Set a minimum number of detected photons in the output distribution
 
         :param value: The minimum photon count
         """
-        self._min_detected_photons = value
+        self._min_detected_photons_filter = value
+
+    def set_min_detected_photons_filter(self, value: int):
```
raksharuia, line 91: Why have this method in both 'noisy_sampling_simulator.py` script and here?

raksharuia (changes requested):

perceval/simulators/simulator.py
ericbrts: We've got two specialized simulators: one for strong simulation, the other for sampling. They both share as much API as possible. For that reason, the same call exists in both classes.

ericbrts (approved):

PR #464, opened 2024-09-18
PCVL-789 Remove deprecated function calls

Benoit-F-Q:


docs/source/notebooks/LOv_rewriting_rules.ipynb
```diff
@@ -698,7 +698,7 @@
     }
    ],
    "source": [
-    "a=pcvl.Circuit.generic_interferometer(4, lambda idx:pcvl.Circuit(2)//pcvl.PS(phi=random.random())//pcvl.BS(theta=random.random()), depth=8, shape=\"rectangle\")\n",
+    "a=pcvl.Circuit.GenericInterferometer(4, lambda idx:pcvl.Circuit(2)//pcvl.PS(phi=random.random())//pcvl.BS(theta=random.random()), depth=8, shape=pcvl.InterferometerShape.RECTANGLE)\n",
```
ericbrts, line 701: It should be `pcvl.GenericInterferometer`.
`Circuit` mistakenly stayed in the call.

tests/test_matrix.py
```diff
@@ -189,8 +189,7 @@ def test_genunitary():
 
 @patch.object(pcvl.utils.logging.ExqaliburLogger, "warn")
```
ericbrts, line 190: You can remove the `@patch` as you're not checking warnings anymore.

PR #469, opened 2024-09-25
PCVL-758 remove deprecated calls from 0.10

ericbrts:


docs/source/legacy.rst
```diff
@@ -2,16 +2,29 @@ Legacy
...
+Circuit.generic_interferometer method was removed
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The :code:`Circuit.generic_interferometer` method has been deprecated since Perceval 0.10.0 and was removed from the
+code base (in order to avoid a circular import).
+
+Please use the :code:`GenericInterferometer` class (from `perceval.components`) directly. See :ref:`Generic Interferometer`
+and you can find a usage example in the :ref:`Circuit Optimizer` code reference.
+
 Breaking changes in Perceval 0.11
```
MarionQuandela, line 22: We should add something about removing the deprecated way of building catalog components

perceval/components/core_catalog/controlled_rotation_gates.py
```diff
@@ -92,7 +92,6 @@ class PostProcessedControlledRotationsItem(CatalogItem):
 
     def __init__(self):
         super().__init__("postprocessed controlled gate")
-        self._default_opts['type'] = AsType.PROCESSOR
```
MarionQuandela, line 95: I just spotted an error in this file, can you change the l104 `if not "n" in kwargs` into `if "n" not in kwargs` ?

MarionQuandela (changes requested):

PR #475, opened 2024-10-29
PCVL-164 detectors

ericbrts:


perceval/components/abstract_processor.py
```diff
@@ -150,11 +153,9 @@ def noise(self, nm: NoiseModel):
     def available_commands(self) -> list[str]:
         pass
 
-    def postprocess_output(self, s: BasicState, keep_herald: bool = False) -> BasicState:
+    def remove_heralded_modes(self, s: BasicState, keep_herald: bool = False) -> BasicState:
         if not keep_herald and self.heralds:
```
Aubaert, line 157: We should remove the keep_herald argument, else it is confusing

perceval/components/abstract_processor.py
```diff
@@ -238,21 +239,33 @@ def add(self, mode_mapping, component, keep_port=True):
...
         elif isinstance(component, AComponent):
             self._add_component(connector.resolve(), component)
         else:
             raise RuntimeError(f"Cannot add {type(component)} object to a Processor")
         self._circuit_changed()
         return self
 
+    def _add_detector(self, mode: int, detector: IDetector):
+        if not isinstance(mode, int):
+            raise TypeError(f"When adding a detector, the mode number must be an integer (got {type(mode)})")
```
Aubaert, line 261: If it's a list of int, can we loop over the values, as iterable are supposed to be supported in `add`?

perceval/components/abstract_processor.py
```diff
@@ -284,6 +297,8 @@ def _compose_processor(self, connector: ModeConnector, processor, keep_port: boo
         # Compute new herald positions
         n_new_heralds = connector.add_heralded_modes(mode_mapping)
         self._n_heralds += n_new_heralds
+        self._mode_type += [ModeType.HERALD] * n_new_heralds
+        self._detectors += [None] * n_new_heralds
```
Aubaert, line 301: So we simply erase the detectors that are on the heralds of the composed processor?

perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -216,6 +207,8 @@ def _noisy_sampling(
             provider: SamplesProvider,
             max_samples: int,
             max_shots: int,
+            detectors: list[IDetector],
+            detection_type: DetectionType,
```
Aubaert, line 211: Seems like detection_type is not used.
Also, why do we give this as argument and not in a function like `set_circuit` and `set_selection`?

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+
+class BSLayeredPPNR(IDetector):
+    """
+    BSLayeredPPNR implements Pseudo Photon Number Resolving detection using layers of beam splitter plugged on
+    2**(number of layers) threshold detectors.
+
+    :param bs_layers: Number of beam splitter layers. Adding more layers enabled to detect
+    """
+
+    def __init__(self, bs_layers: int, reflectivity: float = 0.5):
```
Aubaert, line 73: Do we allow list of reflectivity ?

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+            for m in range(0, 2**(l+1), 2):
+                ppnr_circuit.add(m, BS(BS.r_to_theta(self._r)))
+        return ppnr_circuit
+
+    def detect(self, theoretical_photons: int) -> BSDistribution or BasicState:
+        if theoretical_photons < 2:
+            return BasicState([theoretical_photons])
+
+        if theoretical_photons in self._cache:
+            return self._cache[theoretical_photons]
```
Aubaert, line 108: `self._cache` could be a class instance so we don't need to compute for several instances (in which case we also need to use the number of layers and the reflectivity as a key)

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+
+        from perceval.backends import SLOSBackend
+        ppnr_circuit = self.create_circuit()
+        slos = SLOSBackend()
```
Aubaert, line 112: Given the circuit structure and the input state, we should be able to do this with a basic probability computation, which should save a lot of memory and time

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+        output = BSDistribution()
+        for state, prob in dist.items():
+            state = state.threshold_detection()
+            output[BasicState([state.n])] += prob
+        self._cache[theoretical_photons] = output
+        return output
+
+
+class Detector(IDetector):
+    def __init__(self, p_multiphoton_detection: float = 1):
```
Aubaert, line 126: Here, there is a risk that users use the __init__ method directly instead of the threshold(), pnr() and ppnr() functions. 
Maybe this class should still be abstract and be subclassed by the three detectors ?

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+            return DetectionType.Threshold
+        elif self._pmd == 1:
+            return DetectionType.PNR
+        return DetectionType.PPNR
+
+    def detect(self, theoretical_photons: int) -> BSDistribution or BasicState:
+        if theoretical_photons < 2 or self.type == DetectionType.PNR:
+            return BasicState([theoretical_photons])
+        # Adjust the model to treat the PPNR case here
+        return BasicState([1])
```
Aubaert, line 165: It seems that the PPNR is still to be done here

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+
+
+def detection_type(detectors: list[IDetector]) -> DetectionType:
```
Aubaert, line 168: What the function does doesn't correspond to the docstring

perceval/rendering/circuit/symb_skin.py
```diff
@@ -329,3 +295,9 @@ def port_shape_out(self, port, canvas, mode_style):
         canvas.add_rect((15, 15), 12, 50*port.m - 30, fill="white")
         if port.name:
             canvas.add_text((27, 50*port.m - 9), text='[' + port.name + ']', size=6, ta="right", fontstyle="italic")
+
+    def detector_shape(self, detector, canvas, mode_style):
```
Aubaert, line 299: The shape is always the same for symb skin?

perceval/simulators/_simulate_detectors.py
```diff
@@ -0,0 +1,87 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from perceval.components.detector import IDetector, DetectionType, detection_type
+from perceval.utils import BSDistribution, BasicState
+
+
+def simulate_detectors(dist: BSDistribution, detectors: list[IDetector], min_photons: int = None
```
Aubaert, line 34: Should this function be responsible of the min_photons selection, or should it be done after that (like in the sample case)?

perceval/simulators/simulator_interface.py
```diff
@@ -124,7 +124,7 @@ def probs(self, input_state) -> BSDistribution:
         results, _ = self._postprocess_bsd(results)
         return results
 
-    def probs_svd(self, svd: SVDistribution, progress_callback: callable = None) -> dict:
+    def probs_svd(self, svd: SVDistribution, progress_callback: callable = None, detectors = None) -> dict:
         probs = self._simulator.probs_svd(self._prepare_input(svd), progress_callback)
```
Aubaert, line 128: Missing the detectors transmission here?

perceval/simulators/stepper.py
```diff
@@ -108,7 +108,7 @@ def apply(self, sv: StateVector, r: list[int], c: ACircuit) -> StateVector:
     def probs(self, input_state) -> BSDistribution:
         return _to_bsd(self.evolve(input_state))
 
-    def probs_svd(self, svd: SVDistribution, progress_callback: callable = None) -> dict:
+    def probs_svd(self, svd: SVDistribution, progress_callback: callable = None, detectors = None) -> dict:
```
Aubaert, line 111: Detectors are not used. Should they also be used here so the Stepper simulator gives the same results than the Simulator?

perceval/components/abstract_processor.py
```diff
@@ -238,21 +239,33 @@ def add(self, mode_mapping, component, keep_port=True):
...
+            self.m = component.m + mode_mapping if isinstance(mode_mapping, int) else max(mode_mapping) + 1
             get_logger().debug(f"Number of modes of interest defaulted to {self._n_moi} in processor {self.name}",
-                         channel.general)
+                               channel.general)
 
         connector = ModeConnector(self, component, mode_mapping)
         if isinstance(component, AProcessor):
             self._compose_processor(connector, component, keep_port)
+        elif isinstance(component, IDetector):
+            self._add_detector(mode_mapping, component)
```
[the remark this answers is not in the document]
ericbrts: On L.253

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+        self.name = f"BS-PPNR{bs_layers}"
+        self._layers = bs_layers
+        self._r = reflectivity
+        self._cache = {}  # This cache records simulations for a given photon count to speed up computations
+
+    @property
+    def type(self) -> DetectionType:
+        return DetectionType.PPNR
+
+    def clear_cache(self):
```
ericbrts, line 87: Add docstring

perceval/components/detector.py
```diff
@@ -26,9 +26,160 @@
...
+def detection_type(detectors: list[IDetector]) -> DetectionType:
+    """
+    Computes a global detection type from a given list of detectors.
+
+    :return: PNR if all detectors are PNR or not set
+             Threshold if all detectors are threshold
+             else PPNR
+    """
+    if not detectors:
+        return DetectionType.PNR
```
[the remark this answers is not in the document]
ericbrts: Add a comment here

perceval/rendering/circuit/canvas_renderer.py
```diff
@@ -83,16 +83,24 @@ def open(self):
     def get_circuit_size(self, circuit: ACircuit, recursive: bool = False) -> tuple[int, int]:
         return self._skin.get_size(circuit, recursive)
 
+    def display_input_photons(self, input_pos) -> None:
```
ericbrts, line 86: docstring

perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -206,7 +197,7 @@ def _perfect_samples_no_selection(
 
         return {
             "results": results,
-            "physical_perf": 1,
+            "physical_perf": len(results) / n_samples,
```
ericbrts, line 200: Revert this

perceval/simulators/simulator_interface.py
ericbrts: Will be resolved in PCVL-819

perceval/components/processor.py
```diff
@@ -282,7 +285,7 @@ def probs(self, precision: float = None, progress_callback: callable = None) ->
             self._simulator.set_precision(precision)
         get_logger().info(f"Start a local {'perfect' if self._source.is_perfect() else 'noisy'} strong simulation",
                           channel.general)
-        res = self._simulator.probs_svd(self._inputs_map, progress_callback, self._detectors)
+        res = self._simulator.probs_svd(self._inputs_map, self._detectors, progress_callback)
```
Aubaert, line 288: Why does the sampling function use `set_detectors` and not the prob function?
and this would remove the code breaking

perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -105,6 +105,11 @@ def __init__(self, sampling_backend: ASamplingBackend):
         self._heralds: dict = {}
         self._keep_heralds = True
         self.sleep_between_batches = 0.2  # sleep duration (in s) between two batches of samples
+        self._detectors = None
+
+    def set_detectors(self, detector_list: list[IDetector]):
```
Aubaert, line 110: This should probably be part of the abstract class

perceval/components/processor.py
ericbrts: We discussed yesterday that it made sense to pass the detectors only to `probs_svd` in the strong simulation class, as detectors would be ignored in the majority of the calls (and currently only used in this one - probs_svd). So, I'm still for this way of injecting detectors in the different simulators.

PR #476, opened 2024-11-04
PCVL-580 bridge between heralding & mask

ericbrts:


perceval/simulators/simulator.py
```diff
@@ -392,6 +397,8 @@ def _probs_svd_fast(self, input_dist, p_threshold, progress_callback: Callable |
                 exec_request = progress_callback(progress, 'recombine distributions')
                 if exec_request is not None and 'cancel_requested' in exec_request and exec_request['cancel_requested']:
                     raise RuntimeError("Cancel requested")
+        if self._logical_perf > 0:
```
Aubaert, line 400: Can you add the details of the formula to prove this is the good one in comments? I spent one hour to prove the formula is correct...

Aubaert, line 400: I think you need `self._physical_perf > 0` if you want to avoid division by 0 (anyway, given the definition of the logical perf, it doesn't make sense to define it if physical perf is 0)

perceval/simulators/simulator.py
```diff
@@ -408,28 +415,45 @@ def probs_svd(self, input_dist: SVDistribution, progress_callback: Callable | No
             * physical_perf is the performance computed from the detected photon filter
             * logical_perf is the performance computed from the post-selection
         """
+        self._reset_perf()
+        min_detected_photons = self._min_detected_photons_filter
+        if self._heralds:
+            self._setup_mask_from_heralds()
+            min_detected_photons = max(min_detected_photons, sum(self._heralds.values()))
```
Aubaert, line 422: I think this should raise an Exception or a warning (except if this is for retrocompatibility). We can't know if the given `min_detected_photons` is lower because the user forgot to count the heralds or if they forgot some photons

perceval/simulators/simulator.py
```diff
@@ -516,11 +540,17 @@ def evolve_svd(self,
         if not isinstance(svd, SVDistribution):
             return SVDistribution(self.evolve(svd))
 
+        self._reset_perf()
+        min_detected_photons = self._min_detected_photons_filter
+        if self._heralds:
+            self._setup_mask_from_heralds()
+            min_detected_photons = max(min_detected_photons, sum(self._heralds.values()))
```
Aubaert, line 547: Same

tests/test_backends.py
Aubaert, line 1: We should also benchmark SLOS before and after this PR to confirm that the mask is working

Aubaert (approved):

perceval/backends/_slos.py
```diff
@@ -104,26 +104,28 @@ def compute(self, u, parent_coefs: Matrix = None, mk: int = None):
...
         self._reset()
         self._symb = use_symbolic
-        self._mask_str = mask
-        self._n = n
-        self._mask = None
+        if mask is not None:
+            self.set_mask(mask)
+        if n is not None:
+            get_logger().warn(
+                f"DeprecationWarning: 'n' parameter is now ignored and deprecated. version=0.12", channel.user)
```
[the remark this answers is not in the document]
ericbrts: I refactored and found out the "n" info is not necessary. What's necessary is the photon count in the input state (and which had to be the exact "n" value, so that could be dangerous). After the refactoring, I'm ignoring the value, but still accepting the parameter (with the warning message) - so backward compatibility seems alright.

perceval/simulators/simulator.py
ericbrts: Not sure I've added a mathematical grade proof, but I explained the computation steps in comments.

perceval/simulators/simulator.py
ericbrts: Even if the user forgot to count some photons, increasing the `min_detected_photons` value will give the same correct results, without simulating input states that have to few. Still, you're right, I can show a user warning in this case.

perceval/backends/_naive.py
```diff
@@ -67,5 +67,5 @@ def _compute_submatrix(self, output_state: BasicState) -> np.ndarray:
                 colidx += 1
         return u_st
 
-    def _compute_permanent(self, M: np.ndarray):
-        return xq.permanent_cx(M, n_threads=1)
+    def _compute_permanent(self, m):
```
raksharuia, line 70: why was the typehinting for the argument `m` removed?

perceval/backends/_naive_approx.py
```diff
@@ -45,15 +46,15 @@ def name(self) -> str:
         return "NaiveApprox"
 
     def _compute_permanent(self, M):
-        permanent_with_error = xq.estimate_permanent_cx(M, self._gurvits_iterations, 0)
+        permanent_with_error = xq.estimate_permanent_cx(M, self._gurvits_iterations)
```
raksharuia, line 49: It seems like `M` is being replaced by `m` at the different locations in this file. You missed it here, was this intentional?

perceval/backends/_abstract_backends.py
```diff
@@ -70,17 +73,45 @@ def samples(self, count: int) -> BSSamples:
...
 
     def __init__(self):
         super().__init__()
-        self._cache_iterator = dict()
+        self._cache_iterator: dict = dict()
+        self._masks_str: list[str] | None = None
+        self._mask: xq.FSMask | None = None
+
+    def set_mask(self, masks: str | list[str]):
+        self.clear_mask()
```
raksharuia, line 85: Missing documentation for mask.

perceval/backends/_naive.py
ericbrts: Because it was not accurate enough, and the actual type is hard to express.

MarionQuandela (changes requested):

perceval/simulators/simulator.py
```diff
@@ -438,12 +457,30 @@ def probs_svd(self,
...
         return {'results': res,
                 'physical_perf': physical_perf,
                 'logical_perf': self._logical_perf}
 
+    def _setup_heralds(self):
+        # Set up a mask corresponding to heralds:
+        mask_str = ""
+        for i in range(self._backend._circuit.m):
+            if i in self._heralds:
+                mask_str += f"{self._heralds[i]}"
```
MarionQuandela, line 472: What happens if number of heralds photon is more than 9 ?

perceval/backends/_abstract_backends.py
```diff
@@ -70,17 +73,45 @@ def samples(self, count: int) -> BSSamples:
...
+        for m in masks:
+            m = m.replace("*", " ")
+            assert len(m) == mask_length, "Inconsistent mask lengths"
+        self._masks_str = masks
+        self._init_mask()
+
+    def _init_mask(self):
+        if self._masks_str is not None and self._input_state is not None:
+            instate = self._input_state
+            assert len(self._masks_str[0]) == instate.m, "Mask and input state lengths have to be the same"
```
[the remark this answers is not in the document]
ericbrts: I think the mask has to have the full state size in order to work.

perceval/backends/_mps.py
```diff
@@ -51,10 +48,12 @@ class MPSBackend(AProbAmpliBackend):
     (Phase shifters and Beam Splitters already implemented)
     """
 
-    def __init__(self):
+    def __init__(self, cutoff : int = None):
```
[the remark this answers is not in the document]
ericbrts: Yes but the default version is related to the circuit. We cannot know when instantiating the MPS backend.

perceval/backends/_mps.py
```diff
@@ -51,10 +48,12 @@ class MPSBackend(AProbAmpliBackend):
...
         super().__init__()
         self._s_min = 1e-8  # minimum accepted value for singular values
         self._cutoff = None  # Bond dimension of MPS
+        if cutoff is not None:
+            self.set_cutoff(cutoff)
```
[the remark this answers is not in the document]
ericbrts: I'd like not to break the syntax of old back-ends too much. And a setter method or a property setter is basically the same.

tests/test_backends.py
ericbrts: I had tested manually that the masks were improving memory usage and computation time.

PR #477, opened 2024-11-08
Pcvl 600 job group progress bar

raksharuia:


perceval/runtime/job_group.py
```diff
@@ -222,9 +231,108 @@ def progress(self) -> list:
...
+
+        elif RunningStatus[status_entry] in status_success:
+            return JobCategories.FIN_SUCCESS
+
+        elif RunningStatus[status_entry] in status_other:
+            return JobCategories.FIN_OTHER
+        else:
+            raise ValueError(f"Unspecified status of job in group with value {status_entry}. Cannot categorize")
+
+    def progress(self) -> dict:
```
Aubaert, line 257: The signature has changed --> this should be written somewhere

perceval/runtime/job_group.py
```diff
@@ -27,20 +27,30 @@
...
+from tqdm import tqdm
+
 
 FILE_EXT_JGRP = 'jgrp'
 JGRP_DIR_NAME = "job_group"
 
 DATE_TIME_FORMAT = "%Y%m%d_%H%M%S"
 DEFAULT_MAX_SAMPLES = 10000
 
+class JobCategories(Enum):
```
Aubaert, line 47: Should we do an Enum that "repeats" the information from the RunningStatus Enum?
I think using collections of Running status would make it more consistent

perceval/runtime/job_group.py
```diff
@@ -222,9 +231,108 @@ def progress(self) -> list:
 
         self._write_job_group_to_disk()  # rewrites the group information on disk
 
-        # replace None with 'not_sent' - more readable
-        status_list = [status if status is not None else "not_sent" for status in status_jobs_in_group]
-        return status_list
+        return status_jobs_in_group
+
+    @staticmethod
+    def _map_job_status_category(status_entry: str) -> JobCategories:
```
Aubaert, line 237: Move this to the JobCategories class

perceval/runtime/job_group.py
raksharuia: Technically, this enum is for 'groups of runningstatus' and also includes handling of 'None' (this does not exist in RunningStatus and refers to jobs that were never sent to cloud). 

TBH, I would have preferred a better solution than mine. When you say "collections of RunningStatus" -> do you mean a list/dict/set/something else?

perceval/runtime/job_group.py
raksharuia: The method does return a python dict, I do not understand which signature has changed.

perceval/runtime/job_group.py
raksharuia: Following our discussions, removed the enum class `JobCategories`.

PR #479, opened 2024-11-14
Pcvl 755 distribution tvd function

raksharuia:


perceval/utils/dist_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+    Computes the Total Variation Distance (TVD) between two input BSDistributions.
+
+    :param dist1: First BSDistribution
+    :param dist2: Second BSDistribution
+    :return : total variation distance between the two BSDistributions (value between 0 and 1)
+    """
+    common_bs = set(dist1.keys()).intersection(dist2.keys())
+
+    if not common_bs:
+        raise ValueError('There are no common BasicStates between the two input distributions. '
```
Aubaert, line 45: Should we return 1 (maximum value for TVD) ?

perceval/utils/dist_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+                         'Cannot compute TVD')
+
+    if common_bs != set(dist1.keys()) or common_bs != set(dist2.keys()):
```
Aubaert, line 48: I think that for non-common states, the formula still stands if we consider the prob of absent state to be 0 (this must be checked but I think it's correct)

perceval/utils/dist_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+        get_logger().warn(f"Distributions have mismatched number of states. {len(common_bs)} common states found "
+                          f"and used to compute TVD. Metric maybe innacurate.", channel.user)
```
Aubaert, line 50: Typo "inaccurate" (one 'n', two 'c')

perceval/utils/dist_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from perceval.utils import BSDistribution
+from perceval.utils.logging import get_logger, channel
+
+
+def tvd_dist(dist1: BSDistribution, dist2: BSDistribution) -> float:
```
MarionQuandela, line 34: Missing some TU regarding this method

perceval/utils/dist_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Computes the Total Variation Distance (TVD) between two input BSDistributions.
+
+    :param dist1: First BSDistribution
+    :param dist2: Second BSDistribution
+    :return : total variation distance between the two BSDistributions (value between 0 and 1)
```
MarionQuandela, line 40: What is a return value for closes BSD (more than 0.9 for instance ?) ?
Can you use this method to replace in UT the dirty s*** I did called `assert_bsd_close_enough` and `assert_bsd_close` ?

perceval/utils/dist_metrics.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if not common_bs:
+        raise ValueError('There are no common BasicStates between the two input distributions. '
+                         'Cannot compute TVD')
+
+    if common_bs != set(dist1.keys()) or common_bs != set(dist2.keys()):
+        get_logger().warn(f"Distributions have mismatched number of states. {len(common_bs)} common states found "
+                          f"and used to compute TVD. Metric maybe innacurate.", channel.user)
+
+    tvd = 0.5 * sum(abs(dist1[basicstate]-dist2[basicstate]) for basicstate in common_bs)
```
MarionQuandela, line 52: basic_state instead of basicstate ?

perceval/utils/dist_metrics.py
MarionQuandela, line 34: If you want instead of `dist1` and `dist2` you can call your arguments `dist_lh` and `dist_rh` (left hand / right hand). It enforces the fact that there is only 2 arguments for this method. It's really nothing if you don't want to change (I just don't like number in variable names ^^')

perceval/utils/dist_metrics.py
MarionQuandela, line 45: Basic States instead of BasicStates?

perceval/utils/dist_metrics.py
MarionQuandela, line 50: No need of f string for the second string.

MarionQuandela: And also `inaccurate` instead of `innacurate`

Aubaert (approved):

MarionQuandela (approved):

PR #480, opened 2024-11-18
Pcvl 800 1st feed forward simulator

Aubaert:
Creation of the new feed-forward simulator.

The feed-forward map configurator class serves as an example of configurator.
This PR does not include specific interactions between Processor and FFConfigurator.
These are not the main topic of the PR and should be better implemented in another PR.[https://github.com/Quandela/Perceval/pull/481](url)

### DIFF
--- a/perceval/simulators/feed_forward_simulator.py
+++ b/perceval/simulators/feed_forward_simulator.py
@@ -40,6 +40,7 @@ from .simulator_interface import ISimulator
 class FFSimulator(ISimulator):
 
     def __init__(self, backend: AProbAmpliBackend):
+        super().__init__()
         self._precision = None
         self._heralds = None
         self._postselect = None
@@ -266,6 +267,7 @@ class FFSimulator(ISimulator):
             sim.do_postprocess(self._postprocess)
         else:
             sim.do_postprocess(False)
+        sim.set_silent(True)
         return sim, input_state, detectors + proc.detectors[m:], proc
 
     def _post_process_state(self, bs: BasicState) -> bool:
--- a/perceval/simulators/feed_forward_simulator.py
+++ b/perceval/simulators/feed_forward_simulator.py
@@ -1,0 +1,243 @@
+# MIT License
+#
+# Copyright (c) 2022 Quandela
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# As a special exception, the copyright holders of exqalibur library give you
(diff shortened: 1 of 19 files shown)


tests/test_ff_simulator.py
```diff
@@ -0,0 +1,190 @@
...
+    PostSelect
+from perceval.algorithm import Sampler
+from perceval.components.feed_forward_configurator import CircuitMapFFConfig
+from perceval.simulators.feed_forward_simulator import FFSimulator
+
+backend = SLOSBackend()
+sim = FFSimulator(backend)
+
+cnot = CircuitMapFFConfig(2, 0, Circuit(2)).add_configuration([0, 1], PERM([1, 0]))
+
```
raksharuia, line 42: Maybe you should add a test here that keeping an offset > 1 does not cause errors in computation.

perceval/simulators/feed_forward_simulator.py
```diff
@@ -0,0 +1,296 @@
...
+        proc = Processor(self._backend, input_state.m)
+
+        for r, c in components:
+            proc.add(r, c)
+
+        if filter_states:
+
+            if new_heralds is not None:
+                for r, v in new_heralds.items():
+                    proc.add_herald(r, v)
```
ericbrts, line 235: This call adds heralds on both the input and output of `proc`. I think you want to fix heralds only in the output, which can be achieved with :

```python
proc.add_port(r, Herald(v), PortLocation.OUTPUT)
```

perceval/simulators/feed_forward_simulator.py
```diff
@@ -0,0 +1,296 @@
...
+                    r = self._components[i][0]
+                    # Does not use the list of modes as circuits can have different sizes
+                    components[i] = (config.config_modes(r)[0], sub_circuits[sub_i])
+
+                new_heralds = {i: state[i] for i in measured_modes}
+                sim = self._init_simulator(input_state, components, filter_states=True, new_heralds=new_heralds)
+
+                if is_svd:
+                    new_prog_cb = partial_progress_callable(prog_cb, j / len(default_res), (j + 1) / len(default_res))
+                    sub_res = sim.probs_svd(input_state, detectors, new_prog_cb, normalize=False)["results"]
```
ericbrts, line 157: When the FF configurator injects a `Processor` with heralds, the `input_state` here needs to be adjusted to take the modes covered by heralds into account.

ericbrts (changes requested):

perceval/simulators/feed_forward_simulator.py
```diff
@@ -0,0 +1,296 @@
...
+                'DeprecationWarning: Call with deprecated argument "min_detected_photon_filter", please use "min_detected_photons_filter" instead')
+            min_detected_photons_filter = min_detected_photon_filter
+        if min_detected_photons_filter is not None:
+            self.set_min_detected_photons_filter(min_detected_photons_filter)
+        if postselect is not None:
+            self._postselect = postselect
+        if heralds is not None:
+            self._heralds = heralds
+
+    def _probs_or_svd(self,
```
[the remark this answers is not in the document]
Aubaert: It was changed to `_probs_svd`, and both `probs` and `probs_svd` use it with a SVD now

PR #482, opened 2024-11-21
Pcvl 813 user action methods for different job categories in group

raksharuia:


perceval/runtime/job_group.py
```diff
@@ -376,3 +367,55 @@ def delete_job_groups_date(del_before_date: datetime):
         # delete files
         for f in files_to_del:
             JobGroup.delete_job_group(f)
+
+    def list_successful_jobs(self) -> list[RemoteJob]:
```
Aubaert, line 371: These 4 methods could probably be factorized since only one line changes between them

raksharuia: You are absolutely correct, I completely forgot about it after finishing my dev. Fixed.

PR #489, opened 2024-12-05
Pcvl 599 launch a job group run methods

raksharuia:


perceval/runtime/job_group.py
```diff
@@ -413,8 +400,85 @@ def _recreate_unsent_remote_job(job_entry):
...
+        Launches or reruns jobs in the group on Cloud parallely.
+
+        :param rerun: if True rerun failed jobs or run unsent jobs
+        """
+        for job_entry in self._group_info['job_group_data']:
+            if not rerun and job_entry['status'] is None:
+                remote_job = self._recreate_unsent_remote_job(job_entry)
+                remote_job.execute_async()
+            elif rerun and job_entry['status'] in ['ERROR', 'CANCELED']:
+                print('RE launching jobs in parallel')
```
Aubaert, line 458: Remove this print statement

perceval/runtime/job_group.py
```diff
@@ -413,8 +400,85 @@ def _recreate_unsent_remote_job(job_entry):
...
+    def rerun_failed_sequential(self, delay: int):
+        """
+        Reruns Failed jobs in the group on the Cloud in a sequential manner with a
+        user-specified delay between the completion of one job and the start of the next.
+
+        :param delay: number of seconds to wait between re-launching jobs on cloud
+        """
+        self._launch_jobs_sequential(delay=delay, rerun=True)
+
+    def _launch_jobs_parallel(self, rerun: bool):
```
Aubaert, line 447: This function can be merged with the sequential launch with just a parameter that indicates that it should wait for the job to be complete.

Aubaert (changes requested):

perceval/runtime/job_group.py
raksharuia: oops

Aubaert (approved):
Ok for me now

Aubaert:
Found bugs in tqdm progress bar:
- you are counting jobs at each iteration even if they already participated to the count making the count overpass the maximum of the bar
- the total for each bar is the number of jobs in the job group but it would be more logical if it was only the number of jobs that could be in that category (for example if a job is already successful for example if it was loaded from the disk before launching the tqdm progress, it is added to the total of all bars but it should only be added to the total of the success bar)

PR #501, opened 2024-12-17
Added warning message when using an annotated BasicState as input

Aubaert:


perceval/components/abstract_processor.py
```diff
@@ -644,6 +644,10 @@ def check_input(self, input_state: BasicState):
         expected_input_length = self.m
         assert len(input_state) == expected_input_length, \
             f"Input length not compatible with circuit (expects {expected_input_length}, got {len(input_state)})"
+        if input_state.has_polarization:
+            get_logger().warn("Given input state has polarization, that was dropped (use with_polarized_input instead).")
```
MarionQuandela, line 648: `dropped` is a weird way of saying it. I prefer `ignored` as below

perceval/components/abstract_processor.py
```diff
@@ -644,6 +644,10 @@ def check_input(self, input_state: BasicState):
...
+        elif input_state.has_annotations:
+            get_logger().warn("Given input state has annotations, that were ignored. To use them, consider using a StateVector.")
```
raksharuia, line 650: 'were ignored' implies that annotations have been ignored previously but here we are setting it and will be ignored by the computation. 

I would suggest -  '... will be ignored... '

Same comment for the polirazation.

raksharuia (approved):

PR #502, opened 2024-12-17
Pcvl 824 fix lc simulator number of modes

Aubaert:
Also includes this fix on TD simulator

### DIFF
--- a/perceval/simulators/delay_simulator.py
+++ b/perceval/simulators/delay_simulator.py
@@ -84,8 +84,11 @@ class DelaySimulator(ASimulatorDecorator):
         expanded_input = input_state ** self._depth
         return expanded_input * BasicState([0] * (self._expanded_m - self._depth*self._original_m))
 
-    def _prepare_circuit(self, circuit):
-        self._original_m = _retrieve_mode_count(circuit)
+    def _prepare_circuit(self, circuit, m = None):
+        if m is None:
+            self._original_m = _retrieve_mode_count(circuit)
+        else:
+            self._original_m = m
         expanded_circuit, expanded_mode_count = self._expand_td(circuit)
         self._expanded_m = expanded_mode_count
         return expanded_circuit
--- a/perceval/simulators/delay_simulator.py
+++ b/perceval/simulators/delay_simulator.py
@@ -88,6 +88,7 @@ class DelaySimulator(ASimulatorDecorator):
         if m is None:
             self._original_m = _retrieve_mode_count(circuit)
         else:
+            assert m >= _retrieve_mode_count(circuit), "Given circuit size is too small for the components"
             self._original_m = m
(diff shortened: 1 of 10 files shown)


perceval/simulators/loss_simulator.py
```diff
@@ -38,8 +38,11 @@ class LossSimulator(ASimulatorDecorator):
     def _prepare_input(self, input_state):
         return input_state * BasicState([0] * (self._expanded_m - self._original_m))
 
-    def _prepare_circuit(self, circuit):
-        self._original_m = _retrieve_mode_count(circuit)
+    def _prepare_circuit(self, circuit, m = None):
+        if m is None:
+            self._original_m = _retrieve_mode_count(circuit)
+        else:
+            self._original_m = m
```
raksharuia, line 45: I am unsure, do we need an assert here to verify appropriate m is entered at this stage? Or it won't come to that?

perceval/simulators/delay_simulator.py
```diff
@@ -84,8 +84,11 @@ def _prepare_input(self, input_state):
         expanded_input = input_state ** self._depth
         return expanded_input * BasicState([0] * (self._expanded_m - self._depth*self._original_m))
 
-    def _prepare_circuit(self, circuit):
-        self._original_m = _retrieve_mode_count(circuit)
+    def _prepare_circuit(self, circuit, m = None):
+        if m is None:
+            self._original_m = _retrieve_mode_count(circuit)
+        else:
+            self._original_m = m
```
raksharuia, line 91: same comment as the case for LC (below)

perceval/simulators/simulator_factory.py
```diff
@@ -137,5 +137,6 @@ def build(circuit: ACircuit | Processor | list,
 
         if convert_to_circuit:
             circuit = _unitary_components_to_circuit(circuit, m)
-        simulator.set_circuit(circuit)
+        # m can only be retrieved from a Processor (and useful only in that case)
+        simulator.set_circuit(circuit, m if m else None)
```
raksharuia, line 141: I think `m` will never be `None` here, it is set to 0 at the beginning of this method (l73 has `m=0`)

Aubaert: Yes, but the methods in the set_circuit need `m` to be `None` if it is not defined, whereas the `_unitary_components_to_circuit` method needs m to be 0

ericbrts (approved):

perceval/simulators/simulator_factory.py
MarionQuandela, line 141: I think we need an non-linear optics circuit class that will share the same basic interface with the Circuit class in order to give a circuit object to the set_circuit method that will have the m info in it instead of changing all the calls.

PR #504, opened 2024-12-20
Pcvl 847 simplify svd simulation if basic states in the sv are independant incompatible

Aubaert:


perceval/simulators/_simulator_utils.py
```diff
@@ -85,3 +88,20 @@ def _unitary_components_to_circuit(component_list: list, m: int = 0):
     for r, c in component_list:
         circuit.add(r, c)
     return circuit
+
+
+def _split_by_photon_count(sv: StateVector) -> SVDistribution:
```
ericbrts, line 93: It can be very costly to create 1 SVD per SV in a noisy input state. Indeed, currently the `Source` generates a huge SVD (around 4**n elements) containing SV being only one annotated `FockState` at a time. When iterating on it, you'd create everytime one SVD containing exactly 1 element which is going to be hashed in order to be inserted in the dictionary (the SVD being a dict).

One hash call is fast, but 4**n hash calls, can hinder performances a lot.

I see in *simulator.py* that `_split_by_photon_count` is called everytime so:
Did you benchmark your changes on a regular noisy simulation with 8-10 photons? Is it faster than the current algorithm?

perceval/simulators/simulator.py
```diff
@@ -402,23 +385,32 @@ def _probs_svd_fast(self, input_dist, p_threshold, progress_callback: Callable =
...
+                    has_superposed_states = True
+                if not has_annotations and any(bs.has_annotations for bs in split_sv.keys()):
+                    has_annotations = True
+
         p_threshold = max(global_params['min_p'], max_p * self._rel_precision)
-        trimmed_svd = SVDistribution({state: pr for state, pr in svd.items() if pr > p_threshold})
-        return trimmed_svd, p_threshold, has_superposed_states, has_annotations
+        trimmed_svd = SVDistribution({state: pr for state, pr in trimmed_svd.items() if pr > p_threshold})
+        phys_perf = sum(trimmed_svd.values())  # too low probabilities are integrated into the phys_perf
+        return trimmed_svd, p_threshold, has_superposed_states, has_annotations, phys_perf
```
ericbrts, line 413: Cool to have factored the physical performance computation here

perceval/simulators/_simulator_utils.py
Aubaert: After benchmarking, this version was indeed way worse.
For 9 photons with an imperfect source (brightness = 0.8, indistinguishability = 0.9, g2 = 0.1), for the whole `probs` call (not taking into account the SVD generation):
- From previous version: 8.08 s
- From the reviewed version: 24.3 s
- After upgrading (commit `a4ba7db`): 7.27 s

The new version is now slightly better than the previous one even with no superposed states

perceval/simulators/simulator.py
```diff
@@ -392,24 +392,43 @@ def _preprocess_svd(self, svd: SVDistribution) -> tuple[SVDistribution, float, b
...
-                if not has_annotations and any(bs.has_annotations for bs in split_sv.keys()):
-                    has_annotations = True
-
+            if max(sv.n) >= self._min_detected_photons_filter:
+                max_p = max(p, max_p)
+            else:
+                to_remove.append(sv)
         p_threshold = max(global_params['min_p'], max_p * self._rel_precision)
-        trimmed_svd = SVDistribution({state: pr for state, pr in trimmed_svd.items() if pr > p_threshold})
+        trimmed_svd = SVDistribution({state: pr for state, pr in svd.items() if pr > p_threshold and state not in to_remove})
```
Benoit-F-Q, line 401: On crée 2 SVDistribution qui sont des copies partielles dans cette méthode, j'ai l'impression qu'on pourrait n'en faire qu'une seule.

Cette copie est-elle nécessaire ?
Est-ce qu'on itèrerait plus efficacement sur `{state: pr for state, pr in svd.items() if pr > p_threshold and state not in to_remove}`, 3 lignes plus bas ?

perceval/simulators/simulator.py
```diff
@@ -392,24 +392,43 @@ def _preprocess_svd(self, svd: SVDistribution) -> tuple[SVDistribution, float, b
...
+            for sv, p in to_add.items():
+                trimmed_svd[sv] += p
+                max_p = max(trimmed_svd[sv], max_p)
+
+            p_threshold = max(global_params['min_p'], max_p * self._rel_precision)
+            trimmed_svd = SVDistribution({state: pr for state, pr in trimmed_svd.items() if pr > p_threshold and state not in to_remove})
+
         phys_perf = sum(trimmed_svd.values())  # too low probabilities are integrated into the phys_perf
+
+        for sv in trimmed_svd.keys():
```
Benoit-F-Q, line 425: Si tu arrives à l'écrire en python, `has_superposed_states = any([len(sv) > 1 for sv in trimmed_svd.keys()])` permet de quitter la boucle dès qu'on rencontre un `True`

perceval/simulators/simulator.py
```diff
@@ -392,24 +392,43 @@ def _preprocess_svd(self, svd: SVDistribution) -> tuple[SVDistribution, float, b
...
+            if len(sv) > 1:
+                has_superposed_states = True
+            if not has_annotations and any(bs.has_annotations for bs in sv.keys()):
```
Benoit-F-Q, line 428: Idem, on doit pouvoir utiliser un `any` pour permettre de sortir en avance de cette boucle

Aubaert: Yes ça va un peu plus vite

Benoit-F-Q:
Review OK

PR #507, opened 2025-01-03
Pcvl 855 small simulation upgrades

Aubaert:


perceval/simulators/simulator.py
```diff
@@ -389,13 +389,12 @@ def _probs_svd_fast(self, input_dist, p_threshold, progress_callback: Callable =
 
         """
         2nd step of computing logical performance:
-        After the whole mixed state (input_dist) has been simulated, the missing part of the logical performance needs
-        to be normalized against the physical performance. Indeed, the physical performance is less than 1 for all input
-        states in the input distribution which did not contain enough photons.
-        The more input states were filtered by the physical filter, the heavier the missing part of the logical perf.
+        After the whole mixed state (input_dist) has been simulated,
+        the logical performance is the probability of getting a state that is both physically and logically accepted.
+        To get the right logical perf, we need to divide by the physical perf
```
raksharuia, line 394: What needs to be divided by the physical perf? It is not very clear here. Also replace the word `right` with `correct`.

raksharuia (changes requested):

raksharuia (approved):

PR #508, opened 2025-01-07
Pcvl 860 more clever tensorproduct for noisy state generation

Aubaert:
Some benchmarks on my laptop, with a fully imperfect source, for the function `Processor.with_input`:

To generate a state with 9 photons [1, 1, ...]
Before this PR: 50 seconds
After this PR: 40 seconds

To generate a state with 9 photons in 18 modes [1, 0, 1, 0, ...]
Before this PR: 110 seconds
After this PR: 40 seconds  (the time is now almost independant of empty modes)

To generate a state with 9 photons in the same mode [9]
Before this PR: 160 seconds
After this PR: 38 seconds

The modifications of the tensor product method were also done in the Simulator. Again, the benchmark uses a fully imperfect source, but is now for the function `Processor.probs`, using SLOS.

Using a random unitary, with 7 modes and photons [1, 1... ], no photons filter
Before this PR: 32 seconds
After this PR: 30 seconds
Globally, 6 to 8% speedup using different numbers of photons and modes

Using an empty circuit
Before this PR: 7.9 seconds
After this PR: 8 seconds
Globally, 1 to 2% slowdown

[benchmark_sources.txt](https://github.com/user-attachments/files/18329910/benchmark_sources.txt)
[benchmark sources + probs.txt](https://github.com/user-attachments/files/18329911/benchmark.sources.%2B.probs.txt)

### DIFF
--- a/perceval/utils/statevector.py
+++ b/perceval/utils/statevector.py
@@ -392,15 +392,15 @@ class BSDistribution(ProbabilityDistribution):
                             prob_threshold: float = 0) -> BSDistribution:
         """
         Efficient tensor product for a list of BasicState Distribution.
-         Can modify the distributions in place if merge_modes is False.
+         Can modify the distributions in place if merge_modes is False by adding empty modes.
          Performs `len(distributions) - 1` tensor products
          """
-        if len(distributions) == 0:
-            return BSDistribution()
-
         if any(len(dist) == 0 for dist in distributions):
             get_logger().warn("Empty distribution in tensor product. Ignoring it", channel.user)
             distributions = [dist for dist in distributions if len(dist) > 0]
+
+        if len(distributions) == 0:
+            return BSDistribution()
 
         if not merge_modes:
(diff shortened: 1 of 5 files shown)


perceval/utils/statevector.py
```diff
@@ -249,6 +249,29 @@ def tensor_product(svd1, svd2, prob_threshold: float = 0) -> SVDistribution:
                 new_dist[sv] += proba1 * proba2
         return new_dist
 
+    @staticmethod
+    def list_tensor_product(distributions: list[SVDistribution], prob_threshold: float = 0) -> SVDistribution:
+        """Efficient tensor product for a list of distributions"""
+        if len(distributions) == 0:
+            return SVDistribution()
```
raksharuia, line 256: Just like the warning below, maybe we should have one here that a tensor product with empty list of SVD is asked.

Benoit-F-Q (approved):

PR #509, opened 2025-01-07
Updated Qiskit version to 1.3.1 and removed deprecation warnings

Aubaert:
According to Qiskit documentation, the `operation` and `qubits` attributes of the `CircuitInstruction` class already existed in 0.46.3 https://docs.quantum.ibm.com/api/qiskit/0.46/qiskit.circuit.CircuitInstruction
As such, there should be no breaking change here.


raksharuia (approved):

raksharuia:
There are some notebooks that use Qiskit, could you (@Aubaert) please verify they do not have the warnings?

PR #510, opened 2025-01-08
PCVL-839 Update readme images

Benoit-F-Q:


logo-perceval.png
raksharuia, line 1: Is Perceval's logo also going to be updated with the new Quandela symbol?

Benoit-F-Q: No, but it shares the same issue than Quandela's one :
- it could disappear without notification
- its white color can make it invisible, depending on your theme : 
image
(I shaded somewhat the 2 images)

MarionQuandela: Thoses images should not be at root repository, you can put them in the doc picture folder -> docs/source/_static/img

Benoit-F-Q: Hmm, it's not so simple.

These pictures are not related to the documentation ; someone updating the documentation may silently break the readme again.

I'm reluctant to create a folder for only 2 images.

So leaving them in root directory seemed close to 'good practices', following [boost example](https://github.com/boostorg/boost) for instance

PR #512, opened 2025-01-09
PCVL-793 Compute partial fidelity

Benoit-F-Q:


Aubaert:
Since we now have a method that relies on computing the fidelity on a submatrix, we should change the `fidelity` method to accept this new parameter in the norm that is defined in perceval (path is `utils\algorithms\norm.py`), and also maybe the other metrics in this file.
This would allow the users to get what the optimizer gave them using only the result and the target, as it is the case today.

perceval/utils/algorithms/circuit_optimizer.py
```diff
@@ -88,13 +88,15 @@ def max_eval_per_trial(self, value):
...
-                 template: ACircuit
+                 template: ACircuit,
+                 empty_mode_list: list[int] = []
                  ) -> tuple[ACircuit, float]:
         """
         Optimize a template circuit unitary's fidelity with a target matrix or circuit.
 
         :param target: The target unitary circuit or matrix
         :param template: A circuit with variable parameters (supports only beam splitters and phase shifters)
+        :empty_mode_list: list of the modes without input photon, which are ignored during optmisation as this does not alter the computations
```
Aubaert, line 99: Typo "optmisation" instead of "optimisation" (missing 'i')
Also, I would write "this does not alter the results"

perceval/utils/algorithms/norm.py
```diff
@@ -31,45 +31,65 @@
...
-def frobenius_inner_product(A: np.ndarray, B: np.ndarray) -> float:
+def frobenius_inner_product(A: np.ndarray, B: np.ndarray, skip_colums: list[int] = []) -> float:
     # calculates the inner product associated to Frobenius norm
-    return np.trace(np.dot(np.transpose(np.conjugate(A)), B))
+    C = np.dot(np.transpose(np.conjugate(A)), B)
+    # the generic element of C is C_ij = sum(conj(a_ki).b_kj, k=0..N)
+    # the i-th diagonal element C_ii is the product of i-th column of A and B
+    # C_ii = sum(conj(a_ki).b_ki, k=0..N)
+    # as we want to skip cols from A and B, we omit their indices when computing the trace
+    r = range(C.shape[1])
```
Aubaert, line 91: In general, we prefer not to define the `range` object and use it directly in the for loop, unless it is used several times (which is not the case here)

perceval/utils/algorithms/norm.py
```diff
@@ -31,45 +31,65 @@
...
+    elts = [i for i in r if i not in skip_colums]
+    D = C.diagonal()
+    tr = sum([ D[i] for i in elts ])
+    return sum([ C.diagonal()[i] for i in range(C.shape[1]) if i not in skip_colums ])
```
Aubaert, line 95: You are computing the trace twice (you can just return `tr`), or remove all the previous lines

Aubaert (changes requested):

perceval/utils/algorithms/norm.py
```diff
@@ -31,45 +31,65 @@
...
     :param v: the reference unitary
+    :empty_mode_list: list of input modes that should be ignored during
     :return: real [0-1] float fidelity
     """
-    u_dag = np.transpose(np.conjugate(u))
-    f = abs(np.trace(u_dag @ v)) ** 2 / (u.shape[0] * np.trace(u_dag @ u))
+    n = _count_non_skipped_cols(u.shape[1], skip_colums)
+    if n == 0:
+        return 1.0
+    f = abs(frobenius_inner_product(u, v, skip_colums)) ** 2 / (n * frobenius_inner_product(u, u, skip_colums))
```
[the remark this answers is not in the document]
Benoit-F-Q: You're right, if `u` is unitary, as the docstring suggest it, we should get `frobenius_inner_product(u, u, skip_colums) == n`.

Previous code was written as only `v` were unitary, I'm not sure it was on purpose but I kept this assumption

perceval/utils/algorithms/norm.py
```diff
@@ -31,45 +31,65 @@
...
     r""" Calculate the fidelity of a unitary implementation compared to a reference unitary just comparing
     single input-single output probabilities
 
     :param u: the unitary to evaluate
     :param v: the reference unitary
     :return: real [0-1] float fidelity
     """
-    u_dag = np.transpose(np.conjugate(u))
-    f = np.trace(np.dot(abs(u_dag), abs(v)))/u.shape[0]
+    n = _count_non_skipped_cols(u.C.shape[1], skip_colums)
```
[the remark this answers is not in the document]
Benoit-F-Q: typo

perceval/utils/algorithms/norm.py
```diff
@@ -31,45 +31,65 @@
...
     :param v: the reference unitary
-    :return: real [0-1] float fidelity
+    :return: real distance between matrices
     """
-    return np.linalg.norm(u - v)
+    difference = u - v
+    if skip_colums:
+        return np.sqrt(frobenius_inner_product(difference, difference, skip_colums))
+    else:
+        return np.linalg.norm(u - v)
```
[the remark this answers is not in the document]
Benoit-F-Q: Right.

But I'll keep only the first case, as it covers the second one.

perceval/utils/algorithms/norm.py
Benoit-F-Q: Right

PR #516, opened 2025-01-14
PCVL 843 Input ports should be kept when composing a processor to an empty processor

AliseeNguyen:
…cessor

### DIFF
--- a/perceval/components/abstract_processor.py
+++ b/perceval/components/abstract_processor.py
@@ -376,6 +376,7 @@ class AProcessor(ABC):
         self._components += new_components
 
         # Retrieve ports from the other processor
+        # Output ports
         for port, port_range in processor._out_ports.items():
             port_mode = list(mode_mapping.keys())[list(mode_mapping.values()).index(port_range[0])]
             if isinstance(port, Herald):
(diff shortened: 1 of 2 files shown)


tests/test_processor_composition.py
```diff
@@ -115,6 +115,27 @@ def test_remote_processor_creation(requests_mock):
     rp.add(0, BS())
 
 
+def test_remote_processor_composition_input_ports(requests_mock):
+    ls = LogicalState([0, 0])
+    cnot = catalog['postprocessed cnot'].build_processor()
+    cnot.min_detected_photons_filter(2)
```
Aubaert, line 121: You can remove this line (you don't need it since you don't use `cnot.with_input(ls)`)

tests/test_processor_composition.py
```diff
@@ -115,6 +115,27 @@ def test_remote_processor_creation(requests_mock):
...
+    rp.add(0, cnot)
+    rp.with_input(ls)
+
+    # check that the input ports of the cnot are identical to the remote processor
+    for mode in range(4):
+        cnot_input_port = cnot.get_input_port(mode)
+        rp_input_port = rp.get_input_port(mode)
+        assert cnot_input_port == rp_input_port
+
+        cnot_output_port = cnot.get_output_port(mode)
```
Aubaert, line 134: The ticket only asks for the input ports, so here you are testing something that you didn't change (my bad).
Since there were no tests for the output ports, you can keep it though.

perceval/components/abstract_processor.py
```diff
@@ -397,6 +397,12 @@ def _compose_processor(self, connector: ModeConnector, processor, keep_port: boo
             self._postselect = self._postselect or PostSelect()
             self._postselect.merge(other_postselect)
 
+        # Adding input ports from component processor
```
MarionQuandela, line 400: Maybe add this piece of code before or after the _out_ports  piece of code. 
Also, consider breaking this method in sub methods (one for ports & post selection and one for permutation at least ?)

PR #517, opened 2025-01-14
Adding two methods to (coarse grain) simplify a SV distribution

AliseeNguyen:


perceval/utils/statevector.py
```diff
@@ -225,6 +225,58 @@ def sample(self, count: int, non_null: bool = True) -> list[StateVector]:
         results = random.choices(states, k=count, weights=probs)
         return list(results)
 
+    def photon_threshold_simplification(self, photon_threshold):
```
Aubaert, line 228: Missing type of `photon_threshold`

perceval/utils/statevector.py
```diff
@@ -225,6 +225,58 @@ def sample(self, count: int, non_null: bool = True) -> list[StateVector]:
...
+                        sv_list.append(photon_threshold) # apply threshold for this mode
+                    else:
+                        sv_list.append(component[mode])
+            sv = StateVector(sv_list)
+            if sv in simplified_distribution:
+                simplified_distribution[sv] = simplified_distribution[sv] + p
+            else:
+                simplified_distribution[sv] = p
+
+        return SVDistribution(simplified_distribution)
```
Aubaert, line 250: I'm not sure using these methods on a SVDistribution makes sense. 
A SVD is a collection of StateVectors. 
As such, what happens to the superposition if you apply this to a state like `|2, 1> + i |1, 2>` ?
Currently, the coefficients are lost (since you don't use them when you iterate over the sv).
More globally, I think these methods should only be defined for BSDistributions

Aubaert (changes requested):

perceval/utils/statevector.py
raksharuia: I agree.

perceval/utils/statevector.py
```diff
@@ -418,3 +366,52 @@ def tensor_product(bsd1: BSDistribution,
...
         return self._m
+
+    def photon_threshold_simplification(self, photon_threshold: int):
+        r""" Simplify this `BSDistribution` with a photon threshold for each mode
+        (ex: |0,3,0,0> -> |0,1,0,0> if photon_threshold=1)
+
+        :param photon_threshold: the maximum number of photons per mode
+        :return: the simplified distribution
+        """
+        simplified_distribution = {}
```
raksharuia, line 377: `simplified_distribution`can be initialized as an empty `BSDistribution` (it is a descendent of a default dict). You can use `bsd.add()` to populate it within the loop.

perceval/utils/statevector.py
```diff
@@ -418,3 +366,52 @@ def tensor_product(bsd1: BSDistribution,
...
+        return BSDistribution(simplified_distribution)
+
+    def group_modes_simplification(self, group_size: int):
+        r""" Simplify this `BSDistribution` by grouping modes
+        (ex: |1,3,0,0> -> |4,0> if group_size=2)
+
+        :param group_size: the size of the groups of modes
+        :return: the simplified distribution
+        """
+        simplified_distribution = {}
```
raksharuia, line 400: SAme comment as the previous.

perceval/utils/statevector.py
```diff
@@ -366,3 +366,54 @@ def tensor_product(bsd1: BSDistribution,
...
+        r""" Simplify this `BSDistribution` with a photon threshold for each mode
+        (ex: |0,3,0,0> -> |0,1,0,0> if photon_threshold=1)
+        These "coarse grain" simplification methods can be used to decrease the number of components of a given distribution.
+
+        :param photon_threshold: the maximum number of photons per mode
+        :return: the simplified distribution
+        """
+        simplified_distribution = BSDistribution()
+        for bs, p in self.items():
+            bs_list = []
```
Benoit-F-Q, line 380: The pythonic way to compute `bs_list` would be using 'comprehension lists':
```
bs_list = [ min(t,1) for t in bs]
```

perceval/utils/statevector.py
```diff
@@ -366,3 +366,54 @@ def tensor_product(bsd1: BSDistribution,
...
+            for mode in range(len(bs)):
+                if bs[mode] > photon_threshold:
+                    bs_list.append(photon_threshold) # apply threshold for this mode
+                else:
+                    bs_list.append(bs[mode])
+            bs = BasicState(bs_list)
+            if bs in simplified_distribution:
```
Benoit-F-Q, line 387: comme `BSDistribution` dérive de `defaultdict(float)` (et pas de `dict`), on peut utiliser une nouvelle clé directement, comme en C++.

Autrement dit, on peut utiliser `simplified_distribution.add(bs, p)` dans les 2 cas

perceval/utils/statevector.py
Benoit-F-Q: Notice that we bunch states and use FockStates to label bunches, but these labels do not really make sense.

As our aim is to compared distributions, we could provide ways to create bunches for SVD too

perceval/utils/statevector.py
```diff
@@ -366,3 +366,54 @@ def tensor_product(bsd1: BSDistribution,
...
+        These "coarse grain" simplification methods can be used to decrease the number of components of a given distribution.
+
+        :param group_size: the size of the groups of modes
+        :return: the simplified distribution
+        """
+        simplified_distribution = BSDistribution()
+        for bs, p in self.items():
+            bs_list = []
+            i, mode_sum = 0, 0
+            for mode in range(len(bs)):
```
Aubaert, line 406: You can directly iterate on the values of the BS
`for count in bs:`

perceval/utils/statevector.py
```diff
@@ -401,19 +391,6 @@ def group_modes_simplification(self, group_size: int) -> BSDistribution:
...
-                i += 1
-                mode_sum += bs[mode]
-            bs_list.append(mode_sum) # add the last group
-            bs = BasicState(bs_list)
-            if bs in simplified_distribution:
-                simplified_distribution[bs] = simplified_distribution[bs] + p
-            else:
-                simplified_distribution.add(bs, p)
-
+            bs = BasicState([sum([s for s in bs[group_size*k:group_size*(k+1)]]) for k in range(-(len(bs)//-group_size))]) # group modes by groups of size 'group_size'
```
Benoit-F-Q, line 394: `sum([s for s in bs[group_size*k:group_size*(k+1)]])` se simplifie encore en `sum(bs[group_size*k:group_size*(k+1)])`

Aubaert: Can you add a comment to explain that the integer division here performs a ceil operation as this is not standard ?

Aubaert: Also, have you checked what happens if the group size doesn't divide the number of modes ? I'm not sure the splicing here would work

perceval/utils/statevector.py
```diff
@@ -377,18 +377,8 @@ def photon_threshold_simplification(self, photon_threshold: int) -> BSDistributi
...
-                    bs_list.append(photon_threshold) # apply threshold for this mode
-                else:
-                    bs_list.append(bs[mode])
-            bs = BasicState(bs_list)
-            if bs in simplified_distribution:
-                simplified_distribution[bs] = simplified_distribution[bs] + p
-            else:
-                simplified_distribution.add(bs, p)
-
+            bs = BasicState([min(s, photon_threshold) for s in bs]) # for each mode, keep at max 'photon_threshold' photons
```
Aubaert, line 380: In the comment, "keep at most" would be better (yes I know I'm tatillon)

Aubaert (approved):

PR #520, opened 2025-01-17
PCVL-874 Fix bug in doc example for FF configurators

AliseeNguyen:


docs/source/reference/feed_forward_configurator.rst
```diff
@@ -24,9 +24,11 @@ This class directly links measurements to circuits or processors.
 Any circuit or processor matching the default circuit size can be used given all parameters have numerical values.
 
 >>> import perceval as pcvl
+>>> from perceval.components import Detector
```
Aubaert, line 27: I think it's better to keep these examples small, so you should write `pcvl.Detector.threshold` rather than adding an import line

Aubaert (approved):

PR #522, opened 2025-01-23
Pcvl 871 fix test

Benoit-F-Q:


Benoit-F-Q (commented):
OK but I can't "Approve"

tests/_test_utils.py
```diff
@@ -220,3 +221,46 @@ def set_expected_log_number(self, expected_log_number: int):
         :param expected_log_number: expected log number
         """
         self._expected_log_number = expected_log_number
+
+
+def retry(exception_to_check, tries=4, delay=3, backoff=2, logger=None):
```
Aubaert, line 226: For unit tests, we should use a default delay of 0, and a backoff of 1

tests/_test_utils.py
```diff
@@ -220,3 +221,46 @@ def set_expected_log_number(self, expected_log_number: int):
...
+    def deco_retry(f):
+
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            mtries, mdelay = tries, delay
+            while mtries > 1:
+                try:
+                    return f(*args, **kwargs)
+                except exception_to_check as e:
+                    msg = "%s, Retrying in %d seconds..." % (str(e), mdelay)
```
Aubaert, line 254: For unit tests, we should not print a message

ericbrts: It may be better to print than lose the info of what triggered the retry (?)

ericbrts: Replaced print with a warning

PR #524, opened 2025-01-27
Pcvl 864 source sampling

AliseeNguyen:


perceval/components/source.py
```diff
@@ -201,16 +181,48 @@ def generate_distribution(self, expected_input: BasicState, prob_threshold: floa
...
+
+        distributions = [self.probability_distribution(photon_count, prob_threshold) for photon_count in expected_input]
+        dist = SVDistribution.list_tensor_product(distributions, prob_threshold=prob_threshold)
+
         dist.normalize()
         if self.simplify_distribution and self.partially_distinguishable:
             dist = anonymize_annotations(dist, annot_tag='_')
         return dist
 
+    def generate_samples(self, max_samples, expected_input: BasicState):
```
Aubaert, line 195: Can you add the typing for `max_samples` and the returned type ?

perceval/components/source.py
```diff
@@ -201,16 +181,48 @@ def generate_distribution(self, expected_input: BasicState, prob_threshold: floa
...
+        """
```
Aubaert, line 196: It seems the doctring was copied from another function and does not correspond to this function

perceval/components/source.py
```diff
@@ -201,16 +181,48 @@ def generate_distribution(self, expected_input: BasicState, prob_threshold: floa
...
+        :param expected_input: Expected input BasicState
+            The properties of the source will alter the input state. A perfect source always delivers the expected state
+            as an input. Imperfect ones won't.
+        """
+
+        bsd = self._generate_one_photon_distribution()
+
+        samples = []
+        for photon_count in expected_input:
+            if photon_count == 0 or self.is_perfect():
```
Aubaert, line 210: If the source is perfect, you can return before entering this whole logic

perceval/components/source.py
```diff
@@ -201,16 +181,48 @@ def generate_distribution(self, expected_input: BasicState, prob_threshold: floa
...
+                new_samples = [BasicState([photon_count])] * max_samples
+            else:
+                new_samples = bsd.sample(max_samples, non_null=False)
+                for _ in range(photon_count-1):
+                    new_samples_one_mode = bsd.sample(max_samples, non_null=False)
```
Aubaert, line 215: You need to generate the bsd for each input photon since il will change the annotations. If you want, you don't need to generate it each time `if not self.partially_distinguishable`

perceval/components/source.py
```diff
@@ -201,16 +181,48 @@ def generate_distribution(self, expected_input: BasicState, prob_threshold: floa
...
+                    for i in range(len(new_samples_one_mode)):
+                        new_samples[i] = new_samples[i].merge(new_samples_one_mode[i])
+            if len(samples) == 0:
+                samples = new_samples
+                continue
+            for i in range(len(new_samples)):
+                samples[i] = samples[i]*new_samples[i]
```
Aubaert, line 222: Use `*=` for readability

tests/test_source.py
```diff
@@ -110,3 +110,37 @@ def test_source_multiple_photons_per_mode():
     s = Source(emission_probability=ep)
     svd = s.probability_distribution(2)
     assert_svd_close(svd, dict2svd({"|0>": (1-ep)**2, "|1>": ep*(1-ep)*2, "|2>": ep**2}))
+
+
+def test_source_sample():
+    from perceval.utils import BasicState
```
Aubaert, line 116: Put the import at the top of the file

tests/test_source.py
```diff
@@ -110,3 +110,37 @@ def test_source_multiple_photons_per_mode():
...
+
+    nb_samples = 200
+    source = Source(emission_probability=0.9, multiphoton_component=0.1, losses=0.1, indistinguishability=0.9)
+
+    for bs in [BasicState("|1,1>"), BasicState("|{_:0},{_:1}>")]: # with or without annotations
```
Aubaert, line 121: I'm not sure using annotated input in this method makes sense (to be seen with the others)

perceval/components/source.py
```diff
@@ -201,16 +181,48 @@ def generate_distribution(self, expected_input: BasicState, prob_threshold: floa
...
+
+        return samples
```
Aubaert, line 224: It returns a list, I think it should return a `BSSamples`

tests/test_source.py
```diff
@@ -110,3 +110,37 @@ def test_source_multiple_photons_per_mode():
...
+        # generate samples directly from the source
+        samples_from_source = source.generate_samples(nb_samples, bs)
+        assert len(samples_from_source) == nb_samples
+        samples_count = {sample: samples_from_source.count(sample)/nb_samples for sample in samples_from_source}
```
Aubaert, line 125: You can use a conversion method from `utils.conversion` once `samples_from_source` is a `BSSamples`

tests/test_source.py
```diff
@@ -110,3 +110,37 @@ def test_source_multiple_photons_per_mode():
...
+
+        # compare these samples with complete distribution
+        dist = source.generate_distribution(bs,0)
+        samples_diff = {sample:(count-dist[sample])**2 for sample, count in samples_count.items()}
+        diff = (sum(samples_diff.values()))**0.5
+        assert diff < 0.15
```
Aubaert, line 131: How was this value determined ? Also, you can use the `tvd_dist` method from `utils.dist_metrics`

Aubaert (changes requested):

perceval/components/source.py
Aubaert: You still need to create a new BSD for all subsequent photons in the mode.

Aubaert (approved):

PR #532, opened 2025-02-03
Pcvl 891 do not create temporary in bsd tensor products

Aubaert:


Benoit-F-Q:
Note: you can transform recursivity into a loop - this is a general statement.
It will involve  an array for indices in each distribution, I do not known how to do it efficientently with python (is it even doable?)

Benoit-F-Q (approved):

PR #536, opened 2025-02-10
PCVL-888 Use prob threshold when computing detectors

AliseeNguyen:


Benoit-F-Q (approved):

perceval/simulators/_simulate_detectors.py
```diff
@@ -31,14 +31,15 @@
 from perceval.utils import BSDistribution, BasicState
 
 
-def simulate_detectors(dist: BSDistribution, detectors: list[IDetector], min_photons: int = None
-                       ) -> tuple[BSDistribution, float]:
+def simulate_detectors(dist: BSDistribution, detectors: list[IDetector], min_photons: int = None,
+                       prob_threshold: float = 0) -> tuple[BSDistribution, float]:
```
raksharuia, line 35: It might be a good idea to include `prob_threshold` parameter inside `Stepper`'s simulation of detectors.

AliseeNguyen: We do not have directly access to 'prob_threshold' in the 'Stepper', we will need to '_preprocess_svd' to retrieve it

ericbrts (approved):

raksharuia (approved):

PR #538, opened 2025-02-11
Pcvl 821 factor converters code

raksharuia:


perceval/converters/abstract_converter.py
```diff
@@ -102,7 +104,35 @@ def apply_input_state(self):
     def convert(self, gate_circuit, use_postselection: bool = True) -> Processor:
```
ericbrts, line 104: Shouldn't `convert` not be abstract after the refactoring? Can't it be generic ? And only `_generate_gate_sequence` specific?

perceval/converters/qiskit_converter.py
```diff
@@ -64,49 +57,58 @@ def convert(self, qc, use_postselection: bool = True) -> Processor:
...
-        import qiskit  # this nested import fixes automatic class reference generation
+        # import qiskit  # this nested import fixes automatic class reference generation
 
         get_logger().info(f"Convert qiskit.QuantumCircuit ({qc.num_qubits} qubits, {len(qc.data)} operations) to processor",
                     channel.general)
 
-        gate_sequence = _get_gate_sequence(qc)
-        optimized_gate_sequence = label_cnots_in_gate_sequence(gate_sequence)
+        # some limitation in the conversion, in particular measure
+        assert all(isinstance(instruction.operation, self._qiskit.circuit.gate.Gate)
```
ericbrts, line 66: I think this assert should be moved inside `_get_gate_sequence`

perceval/converters/abstract_converter.py
```diff
@@ -77,15 +79,24 @@ def __init__(self, backend_name: str = "SLOS", source: Source = None, noise_mode
     def count_qubits(self, gate_circuit) -> int:
         pass
 
+    @abstractmethod
+    def _get_gate_sequence(self, gate_circuit) -> list[list]:
+        pass
```
ericbrts, line 84: Could you write a small docstring instead of `pass`, explaining what's expected from the implemented methods?

perceval/converters/abstract_converter.py
```diff
@@ -77,15 +79,24 @@ def __init__(self, backend_name: str = "SLOS", source: Source = None, noise_mode
...
+
+    @abstractmethod
+    def _get_qubit_names(self, gate_circuit):
+        pass
+
+    # @abstractmethod
```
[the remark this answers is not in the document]
raksharuia: I missed to remove it earlier, done.

perceval/converters/abstract_converter.py
```diff
@@ -77,15 +79,24 @@ def __init__(self, backend_name: str = "SLOS", source: Source = None, noise_mode
     def count_qubits(self, gate_circuit) -> int:
         pass
 
+    @abstractmethod
+    def _get_gate_sequence(self, gate_circuit) -> list[list]:
+        pass
+
+    @abstractmethod
+    def _get_qubit_names(self, gate_circuit):
+        pass
```
ericbrts, line 88: Same: docstring instead of `pass`

perceval/converters/cqasm_converter.py
```diff
@@ -178,34 +173,14 @@ def convert(self, ast, use_postselection: bool = True) -> Processor:
         :return: the converted processor
         """
         if isinstance(ast, str):
-            return self._convert_from_string(ast, use_postselection)
+            ast = self._convert_from_string(ast, use_postselection)
 
         get_logger().info(f"Convert cqasm.ast ({len(self._qubit_list)} qubits, {len(ast.block.statements)} operations) to processor",
                     channel.general)
         self._collect_qubit_list(ast)
         self._use_postselection = use_postselection
```
ericbrts, line 181: Is `self._use_postselection` still required? I think not, so it can be removed

perceval/converters/abstract_converter.py
```diff
@@ -77,15 +79,24 @@ def __init__(self, backend_name: str = "SLOS", source: Source = None, noise_mode
     def count_qubits(self, gate_circuit) -> int:
         pass
 
+    @abstractmethod
+    def _get_gate_sequence(self, gate_circuit) -> list[list]:
+        pass
+
+    @abstractmethod
+    def _get_qubit_names(self, gate_circuit):
```
ericbrts, line 87: Seems you need `n_qubits` as additional parameter in `_get_qubit_names` as it's used in the QiskitConverter

raksharuia: Yes, you are correct. My pycharm did not even complain about it and I missed to add the parameter here.

perceval/converters/cqasm_converter.py
raksharuia: Nope, not really.

PR #541, opened 2025-02-14
PCVL 890 Embed images in ReadMe.md for pypi publication

Benoit-F-Q:


setup.py
```diff
@@ -1,9 +1,51 @@
...
+import sys
 import setuptools
 from os.path import dirname
 from glob import glob
 
+
+def getLocalFileContent(filename):
+    if not os.path.isfile(filename):
+        return None
+    fileExtension = filename.split('.')[1].lower()
```
AliseeNguyen, line 14: You should check that '.' is in `filename`, the regex used don't check that

AliseeNguyen (approved):

PR #545, opened 2025-02-20
PCVL-868 Set min detected photons with logical state

AliseeNguyen:


Aubaert (approved):

perceval/components/abstract_processor.py
```diff
@@ -643,6 +643,7 @@ def heralds(self):
 
     def _with_logical_input(self, input_state: LogicalState):
         input_state = get_basic_state_from_ports(list(self._in_ports.keys()), input_state)
+        self._min_detected_photons_filter = input_state.n + list(self.heralds.values()).count(1)
```
ericbrts, line 646: I think this is overriding the user value that could exist. I'm not sure that's the behaviour we want. And if we want it, we should warn the user that his given value has been overridden.
What do you think?

Aubaert: Yes, I think we should check that the value is not None before

PR #547, opened 2025-02-20
Pcvl 865 use source sampling in sampling simulator

Aubaert:


perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -293,21 +315,10 @@ def _noisy_sampling(
...
-                                not_selected_physical += 1
-                            elif state.n < self._min_detected_photons_filter:
-                                not_selected_physical += 1
-                                shots += 1
-                                if max_shots is not None and shots + len(selected_inputs) >= max_shots:
-                                    break
-                            else:
-                                # Must be considered as already shot, but will be counted later
-                                selected_inputs.append(state)
+                        selected_inputs, new_rejected, shots = self._filter_generated_inputs(generated, shots, max_shots)
```
Benoit-F-Q, line 321: `_filter_generated_inputs` ne devrait prendre que `shots` (qui serait donc coupé à `max_shots`).

Cela permet de tester explicitement que `shots < max_shots` plutôt que `len(selected_inputs)` un peu plus bas (même si le commentaire aide)

perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -68,28 +84,81 @@ def prepare(self, noisy_input: BSDistribution, n_samples: int, progress_callback
             self._weights[input_state] = math.ceil(0.1 * self._weights[input_state])
 
             if progress_callback:
-                cancel_request = progress_callback(0, 'prepare')
+                cancel_request = progress_callback(0, 'prepare sampling')
```
Benoit-F-Q, line 87: Je ne comprends pas la présence de cette callback (et surtout du sleep) dans la boucle.

(Ce n'est pas le développement en cours, mais c'est l'occasion de poser la question).

perceval/simulators/noisy_sampling_simulator.py
Aubaert: The thing here is that we can reach `max_shots` during the generation process (in which case it becomes useless to generate further) but there can still be useful states, in which case `selected_inputs` is not empty.

perceval/simulators/noisy_sampling_simulator.py
ericbrts: One feature Perceval offers, is to be able to be informed of a (very) long algorithm progress and customize what's happening during this progress update. That's why progress callbacks exist in Perceval.

The sleep is a work-around for a specific use of this in a specific context, we should look into it, some day.

perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -321,39 +431,59 @@ def samples(self,
...
 
-        new_input, pre_physical_perf = self._preprocess_input_state(svd, max_p, prepare_samples)
-
-        # Prepare pools of pre-computed samples
         provider = SamplesProvider(self._backend)
         provider.sleep_between_batches = self.sleep_between_batches
-        provider.prepare(new_input, prepare_samples, progress_callback)
 
-        res = self._noisy_sampling(new_input, provider, max_samples, max_shots, det_type, progress_callback)
+        if source_defined:
```
ericbrts, line 456: Would it be possible to split this long methods into steps, to make it clearer. This if / else is hard to understand here in the middle of the code.

ericbrts (approved):

PR #548, opened 2025-02-24
PCVL-873 warning not to use KLM cnot should be moved to the catalog doc

AliseeNguyen:


perceval/components/core_catalog/klm_cnot.py
```diff
@@ -41,7 +41,7 @@
 
 class KLMCnotItem(CatalogItem):
     article_ref = "https://doi.org/10.1073/pnas.1018839108"
-    description = f"{_GATE_NAME} gate with 4 ancillary modes\n{_WARNING_NOT_OPTIMAL}"
+    description = f"{_GATE_NAME} gate with 4 ancillary modes.\n{_WARNING_NOT_OPTIMAL}"
```
Benoit-F-Q, line 44: Merge `_WARNING_NOT_OPTIMAL` with `description` as the warning string isn't use elsewhere anymore

Benoit-F-Q (approved):

PR #550, opened 2025-02-24
PCVL-866 Parameter expressions

anthonyrtw:
#### Updates to `Parameter` Class:
- Introduced arithmetic magic methods, enabling arithmetic operations on `Parameter` objects to return `Expression`  instances.
- Introduced observer pattern to update any dependent expressions:
  - `_add_expression`: Called when an `Expression` containing a `Parameter` is created.
  - `_notify_expressions`: Called when `Parameter.set_value()` is invoked ensuring dependent `Expression` objects update accordingly.

```python
A = Parameter("A")
B = Parameter("B")
expr = 2 * A + B
print(expr)
print(expr.parameters, '\n')

A.set_value(2) 
B.set_value(1)
print(expr)
print(expr.parameters)
```
This should output:
```python
Expression(2 * A + B, value=None) 
[Parameter(name='A', value=None), Parameter(name='B', value=None)]

Expression(2 * A + B, value=5) 
[Parameter(name='A', value=2), Parameter(name='B', value=1)]
```

#### Updates to `Expression` Class:
- Modified `Expression` to accept two arguments:
  - `name`: A string representing the equation.
  - `parameters`: A set containing all parameters present in `name`.
- If a parameter in `name` is not explicitly provided in `parameters`, it is automatically created. 
- The `parameters` attribute stores all parent parameters in a list.

```python 
X, Y = P("X"), P("Y")
expr = Expression("2 * X + Y + Z", {X, Y})  # Parameter Z will be created automatically
print(expr.parameters)
```
```python 
[Parameter(name='X', value=None), Parameter(name='Y', value=None), Parameter(name='Z', value=None)]
```

#### Additional Changes:
- Enabled inversion of undefined parametrized beamsplitters and phase shifters by using Expressions.

```python
A, B = P("A"), P("B")
bs = BS(A + B) 
bs_inv = bs.inverse(h=True) 
pdisplay(bs)
```
image

- Added option to `AParameterizedComponent.get_parameters()` to enable showing Expressions or just return parent Parameters belonging to any Expressions present in the component.

```python 
A, B = P("A"), P("B")
circ = BS(A ** 2 - B) // PS(A)
print(circ.get_parameters()) 
print(circ.get_parameters(expressions=True))
```
```python 
[Parameter(name='B', value=None), Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586)]
[Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586), Expression(A**2 - B, value=None)]
```


perceval/utils/parameter.py
```diff
@@ -212,16 +218,197 @@ def max(self, m: bool):
...
+            return self._original
+        
+        if isinstance(self, Expression):
+            expr = Expression(f"(-{self.name})", self._params)
+        else:
+            expr = Expression(f"(-{self.name})", {self})
+        expr._original = self
+        return expr
 
 class Expression(Parameter):
```
ericbrts, line 298: I suggest we move this class to a new expression.py file, for maintainability reasons.

perceval/utils/parameter.py
```diff
@@ -212,16 +218,197 @@ def max(self, m: bool):
...
-        super().__init__("_%d" % Parameter._id, e, is_expression=True)
+            e = sp.S(name)
+            self.name = f"({e})"
+        except Exception as err:
+            raise ValueError("%s is not an expression: %s", name, str(err))
+        assert isinstance(e, sp.Expr), "%s is not an expression" % name
+        
+        # Create set containing all parent parameters
+        self._params = set() if parameters is None else set(parameters)
+        self._create_missing_parameters()
```
ericbrts, line 317: I'm not sure if that's better to autogenerate missing parameters, or to force the user to create them all (I'm feeling more confident to force the user to create all parameters, that way if they mess up their expression, they'll know it asap - at Expression instantiation - rather than at evaluation. It's better to fail fast).

On top of that, I tried this code:
```python
X, Y = P("X"), P("Y")
expr = E("2 * cos(X) + Y", {X, Y})
print(expr.parameters)

X.set_value(np.pi/4)
Y.set_value(3)
print(float(expr))
print(2*np.cos(np.pi/4) + 3)
```
which fails because of the `self._create_missing_parameters()` call which detects a "cos" parameter and creates it. This "cos" parameter shadows the Sympy cosinus function.
It works well with "2 * cos(X) + Y" as long as I comment out `self._create_missing_parameters()`.

To conclude this long comment, for the 2 reasons above, I'd suggest removing the feature to autogenerate the missing parameters.

perceval/utils/parameter.py
```diff
@@ -135,6 +138,7 @@ def set_value(self, v: float, force: bool = False):
         if self.fixed and not force:
             raise RuntimeError("cannot set fixed parameter", v, self._value)
         self._value = v
+        self._notify_expressions()
```
ericbrts, line 141: That's a cool "observer" pattern you've implemented there. However, I'm wondering if that's not overkill for our needs. On top of that, as it's not totally "abstracted", it looks like a circular dependency (Parameter needs to know about Expression, and Expression needs to know about Parameter).

Alternatively, we could ask `Expression.__float__` to re-evaluate at every call. This would be a problem only if you're unsing the same expression in hundreds of different places at once.

In your wildest use case, in how many places are you using a single expression?

ericbrts (changes requested):
That's a cool contribution!

It's still missing automated tests and documentation + I have a few comments on how `Parameter` and `Expression` classes work together.
We devs can help on the test and documentation part.

Thanks,
Eric

perceval/utils/parameter.py
anthonyrtw: Hi Eric. Thanks for the helpful feedback!

I think you're right regarding the auto-generation of missing parameters. I can take it out.

perceval/utils/parameter.py
```diff
@@ -212,16 +218,197 @@ def max(self, m: bool):
...
+
+    @property
+    def parameters(self) -> list[Parameter]:
+        """Returns list of parent parameters in alphabetical order"""
+        return sorted(self._params, key=lambda obj: obj.name)
+
+    @property
+    def is_periodic(self) -> bool:
+        """Expressions are not considered periodic"""
+        return False
```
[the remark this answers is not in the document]
anthonyrtw: Thanks for the feedback! 

I think I avoided period handling for the Expression class as some elements of this feature are not exactly trivial to determine for generic Expressions. For example, determining the `min_v`, `max_v` values of a generic Expression given the corresponding parent Parameters' values could be tricky.

perceval/utils/parameter.py
```diff
@@ -70,8 +69,8 @@ def __init__(self, name: str, value: float = None,
         self._periodic = periodic
         self._pid = Parameter._id
         self._is_expression = is_expression
```
Aubaert, line 71: This could be set as property, like
```
@property
def _is_expression(self):
    return not type(self) == Parameter
```
This would avoid having to give this as argument

perceval/utils/parameter.py
```diff
@@ -220,186 +216,100 @@ def max(self, m: bool):
...
-        
-    def _update_value(self):
+    def __float__(self):
         """Updates Expression with respect to any changes made to parent Parameters"""
         if any(not param.defined for param in self._params):
-            self._value = None
-        else:
-            self._value = sp.S(self.name).subs({param.name : param._value for param in self._params})
+            raise ValueError("Expression is symbolic, cannot compute its numerical value")
+        return float(self.spv.subs({param.name : param._value for param in self._params}))
```
Aubaert, line 312: Only the basic Parameters are used here.
I'm not sure that we want to keep the Expressions in `self._params`

ericbrts: Not sure I understand your comment. Only `Parameter` are stored in `self._params`

Aubaert: You're right, I thought it was still the case but you changed it

tests/test_serialization.py
```diff
@@ -304,3 +304,28 @@ def test_compress():
...
+
+    assert "A" in c_deser.params and "B" in c_deser.params
+
+    c_deser.param("A").set_value(0.2)
+    c_deser.param("B").set_value(0.8)
+
+    assert float(c_deser._components[1][1].param('theta')) == pytest.approx(0.2 + 0.8)
+
+    u = c_deser.compute_unitary()
+    assert numpy.allclose(u, [[ 0.80830707 + 0.34174675j, -0.09524715 + 0.46986895j],
```
Aubaert, line 330: Why not use `c.compute_unitary()` after setting values for `p_a` and `p_b` ?

Aubaert (approved):

perceval/components/unitary_components.py
```diff
@@ -217,8 +238,8 @@ def describe(self):
 
     def inverse(self, v=False, h=False):
         if h:
-            if self._phi.is_symbolic():
-                raise ValueError('Cannot invert PS with variable parameters')
+            if self._phi.is_symbolic:
+                self._phi = -self._phi
```
anthonyrtw, line 242: I just noticed a bug with the inversion of phase shifters here. The symbolic representation of the phase shifter updates accordingly, but the `_params` attribute and `pdisplay` does not

```python 
from perceval import PS, pdisplay, P

A = P("A")
ps = PS(A)
ps.inverse(h=True)

print('_params:', ps._params, '\n')
pdisplay(ps)
```  
``` 
_params: {'phi': Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586), 'max_error': Parameter(name='max_error', value=0, min_v=0.0, max_v=3.141592653589793)} 
```
image

anthonyrtw:
Hi, I also spotted another bug related to inverting circuits containing parameterized BS gates in the Hadamard basis: 

```python 
from perceval import P, Circuit, Processor, BS

A = P("A")
B = P("B")

circ = Circuit(2)
circ.add(0, BS.H(A))
circ.add(0, BS.H(B))
circ.inverse(h=True)

processor = Processor("SLOS", 2)
processor.set_circuit(circ)
processor.linear_circuit()
```
``` 
RuntimeError: two parameters with the same name in the circuit (theta)
```

This does not seem to happen with `BS.Rx` or `BS.H`.

ericbrts (approved):

perceval/components/unitary_components.py
ericbrts: Fixed

PR #551, opened 2025-02-25
PCVL-875 Wrap text to new line if too long in pdisplay

AliseeNguyen:


perceval/rendering/circuit/phys_skin.py
```diff
@@ -167,7 +167,7 @@ def ffconf_shape(self, comp: AFFConfigurator, canvas, mode_style):
 
         # The actual component
         canvas.add_rect((5, 10), 50 * w - 10, 50 * comp.m - 20, fill="lightgreen")
-        canvas.add_text((w * 25, 30 + 50*(comp.m-1)/2), size=10, ta="middle", text=comp.name)
+        canvas.add_text((w * 25, 30 + 50*(comp.m-1)/2), size=10, ta="middle", text=comp.name, max_size=50*w)
```
Aubaert, line 170: This should be done everywhere we have a box: Unitaries, sub-circuits (maybe others that I'm forgetting)

perceval/rendering/canvas/canvas.py
```diff
@@ -262,6 +262,31 @@ def add_text(self, points: tuple[float, float],
...
+    def normalize_text(self, text, size, points, max_size):
+        if max_size is not None and size * len(text) > max_size:
+            new_size = int(max_size / len(text) * 1.5)
+            if new_size < size / 2:
+                size = size / 2
+                # try to split text on spaces instead of reducing size
+                pos_x, pos_y = points
+                remaining = text
+                new_text = ''
+                min_index = int(max_size/size * 0.8)
```
Aubaert, line 274: The use of a min index here is strange.
For example, if the first space is before the min index and the second is after the max index, the text will be cut in the middle of the intermediate word.
I think it is better to split on spaces and change the line if the next word would make the number of characters above the limit.

perceval/rendering/circuit/phys_skin.py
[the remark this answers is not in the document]
ericbrts: Yes

PR #555, opened 2025-02-26
PCVL-905 Make min_detected_photons_filter mandatory in noisy and remo…

AliseeNguyen:
…te computations


perceval/components/processor.py
```diff
@@ -224,7 +225,8 @@ def with_polarized_input(self, bs: BasicState):
         if 'min_detected_photons' in self._parameters:
             self._min_detected_photons_filter = self._parameters['min_detected_photons']
         if self._min_detected_photons_filter is None:
-            self._deduce_min_detected_photons(bs.n)
+            raise ValueError("The value of min_detected_photons is not set." +
```
ericbrts, line 228: Same, no exception

tests/test_loss_channel.py
```diff
@@ -46,8 +46,8 @@
 
 def test_lc_minimal():
     p = Processor("SLOS", 1).add(0, LC(loss))
-    p.with_input(SVDistribution(BasicState([2])))
     p.min_detected_photons_filter(0)
+    p.with_input(SVDistribution(BasicState([2])))
```
ericbrts, line 50: This test original code should keep working.

ericbrts (changes requested):

Aubaert (changes requested):

Aubaert (approved):

perceval/components/processor.py
```diff
@@ -173,10 +173,8 @@ def with_input(self, input_state: BasicState) -> None:
         The properties of the source will alter the input state. A perfect source always delivers the expected state as
         an input. Imperfect ones won't.
         """
-        if 'min_detected_photons' in self._parameters:
-            self._min_detected_photons_filter = self._parameters['min_detected_photons']
         if self._min_detected_photons_filter is None and self._source.is_perfect():
```
ericbrts, line 176: I think it's not the correct location to check if the simulation is perfect or not. Noise can still be injected after the call to `Processor.with_input`.
In one of yesterday's comment, I suggested this default value in the special case of a perfect simulation is set just before checking that a user value is available (i.e. at the beginning of the `probs` and `samples` methods).
Could you factor this check + special case in a method?

ericbrts (approved):

PR #564, opened 2025-03-05
PCVL-899 proxy config parameters

AliseeNguyen:


perceval/runtime/_token_management.py
```diff
@@ -101,6 +101,7 @@ def force_token(token: str):
         TokenProvider._CACHED_TOKEN = token
 
 
+@deprecated(version="0.13.0", reason="Use instead RemoteConfig class method `save`")
```
Aubaert, line 104: I think the text should say "... class methods `set_token` then `save`"

perceval/runtime/remote_config.py
```diff
@@ -0,0 +1,97 @@
...
+        :return: Current token
+        """
+        return RemoteConfig._token
+
+    @staticmethod
+    def save():
+        """Save the current remote configuration
+        """
+        token_provider = TokenProvider()
+        token_provider.force_token(RemoteConfig._token)
```
Aubaert, line 89: Should the `force_token` be done only when calling `save` ? Also, once we have a `RemoteConfig` that knows the token, should we keep the `TokenProvider` class ?

perceval/runtime/_token_management.py
```diff
@@ -101,6 +101,7 @@ def force_token(token: str):
...
 def save_token(token: str):
     """Save provided token into persistent data, replace any token previously saved
```
Aubaert, line 106: I think this should replace the token in the `RemoteConfig` too, since it's the one that will be used by `RemoteProcessor`

perceval/runtime/_token_management.py
```diff
@@ -109,3 +111,5 @@ def save_token(token: str):
     token_provider = TokenProvider()
     token_provider.force_token(token)
     token_provider.save_token()
+
+    pcvl.runtime.remote_config.RemoteConfig.set_token(token)
```
Aubaert, line 115: This should also save it as suggested by the methods name (and was the previous behaviour)

AliseeNguyen: It is already saved with `token_provider.save_token()`. If we call `RemoteConfig.save()`, it will also save the proxies.

perceval/runtime/_token_management.py
```diff
@@ -30,10 +30,13 @@
 
 import os
 
-from perceval.utils.logging import get_logger, channel
+import perceval as pcvl
```
Aubaert, line 33: Its better to avoid importing everything

perceval/runtime/_token_management.py
```diff
@@ -71,13 +77,15 @@ def _from_file(self) -> str | None:
...
 
+    @deprecated(version="0.13.0", reason=f"Use instead RemoteConfig class method `get_token`")
     def get_token(self) -> str | None:
         """Search for a token to provide
 
         :return: A token, or None if no token was found
         """
         return TokenProvider._CACHED_TOKEN or self._from_environment_variable() or self._from_file()
 
+    @deprecated(version="0.13.0", reason=f"Use instead RemoteConfig class methods `set_token` then `save`")
```
Aubaert, line 88: I think you can safely remove all the private methods (starting with _ ). For the other methods, you should make them continue to work, so you should call the methods from `RemoteConfig` internally, as it's the one that is called in the `RemoteProcessor`. The `TokenProvider` should no longer do anything on its own

tests/test_token_provider.py
Aubaert, line 1: The `TokenProvider` is only deprecated, but it should still work. I'm not sure you can delete the test file now

perceval/runtime/_token_management.py
```diff
@@ -46,61 +43,64 @@ class TokenProvider:
     - File on the disk
     """
 
-    _CACHED_TOKEN = None
-
     def __init__(self, env_var: str = "PCVL_CLOUD_TOKEN"):
```
ericbrts, line 46: I would say that the __init__ of the TokenProvider should be marked "deprecated", not its methods.
Nobody uses it anymore in our code, I guess. Or are there still a usage somewhere?

perceval/runtime/rpc_handler.py
```diff
@@ -50,7 +50,7 @@ class RPCHandler:
 
     """
 
-    def __init__(self, name, url, token):
```
ericbrts, line 53: There's another implementation of `RPCHandler` in `perceval.providers.scaleway`. Could you make the same changes there, please (if it's easy)?

ericbrts (approved):

PR #571, opened 2025-03-19
PCVL-886 Processors mixing different Encoding on same modes

AliseeNguyen:


raksharuia (approved):

perceval/components/experiment.py
```diff
@@ -400,6 +400,16 @@ def _compose_experiment(self, connector: ModeConnector, experiment: Experiment,
             if self.are_modes_free(range(port_mode, port_mode + port.m), PortLocation.INPUT):
                 self.add_port(port_mode, port, PortLocation.INPUT)
 
+        # Check port composition
+        for m_out, m_in in mode_mapping.items():
+            out_port = self.get_output_port(m_out)
+            in_port = experiment.get_input_port(m_in)
+            if out_port is not None and in_port is not None and out_port.encoding != in_port.encoding:
```
Aubaert, line 407: This should also check that the ports are entirely on the same modes (e.g. if you have a dual rail on modes 0 and 1 for the left hand side, and a dual rail on modes 1 and 2 for the right hand side)

ericbrts (approved):

PR #575, opened 2025-03-24
PCVL-924 Avoid PersistentData creating folders and files when not needed

AliseeNguyen:


perceval/utils/logging/config.py
```diff
@@ -43,10 +43,10 @@ class LoggerConfig(dict):
     """This class represent the logger configuration as a dictionary and can be used to save it into persistent data.
     On class initialization, the configuration will be loaded from persistent data.
     """
-    def __init__(self):
+    def __init__(self, persistent_data: PersistentData = PersistentData()):
```
Aubaert, line 46: Prefer passing `None` by default and initializing the object after, here it is initialized when the method is defined, even if it's never used

Aubaert (approved):

perceval/runtime/_token_management.py
```diff
@@ -47,7 +47,7 @@ class TokenProvider:
     def __init__(self, env_var: str = "PCVL_CLOUD_TOKEN", remote_config: RemoteConfig = RemoteConfig()):
         """
         :param env_var: Environment variable name to search for a token (default PCVL_CLOUD_TOKEN)
-        :param file_path: Path to search for a file containing a token (default None)
+        :param remote_config: The associated remote config
```
raksharuia, line 50: It might be better to say `The associated remote configuration` in the description.

raksharuia (approved):

perceval/utils/logging/loggers.py
```diff
@@ -132,9 +122,13 @@ def apply_config(self, config: LoggerConfig):
         self._configure_logger(config)
 
     def get_log_file_path(self):
-        return path.join(PersistentData().directory, "logs", "perceval.log")
+        return path.join(self._persistent_data.directory, "logs", "perceval.log")
 
     def enable_file(self):
+        if self._persistent_data.is_writable():
```
ericbrts, line 128: From what I understand, Exqalibur logging system is only initialised when calling `enable_file()`.
I could never call this method and still expect logging events to happen.

For instance, without your PR,
```python
import perceval as pcvl
from perceval.utils.logging import level, channel

pcvl.use_perceval_logger()

pcvl.get_logger().error("AAA")

l = pcvl.get_logger()
l.set_level(level.info, channel.user)
l.set_level(level.debug, channel.general)
l.set_level(level.warn, channel.resources)

l.info("INfo USER", channel.user)
l.error("BBB")
l.info("INfo RESOURCES", channel.resources)
l.warn("Warn")
```
Ouputs:
[info] === Starting Perceval session - process ID: 79f3c3b5-c30d-4dd6-8dc2-00c6c9e94794 ===
[error] AAA
[info] Set log level to 'info' for channel 'user'
[info] Set log level to 'debug' for channel 'general'
[info] Set log level to 'warn' for channel 'resources'
[info] INfo USER
[error] BBB
[warning] Warn

With your branch, it outputs nothing.

My config.json looks like that:

```json
{
    "logging": {
        "use_python_logger": false,
        "enable_file": false,
        "channels": {
            "general": {
                "level": "info"
            },
            "resources": {
                "level": "off"
            },
            "user": {
                "level": "warn"
            }
        }
    }
}
```

PR #576, opened 2025-03-26
PCVL-799 FFConfigurator serialisation

ericbrts:


perceval/serialization/_circuit_serialization.py
```diff
@@ -138,6 +139,51 @@ def _serialize(self, barrier: comp.Barrier):
         pb_barrier.visible = barrier.visible
         self._pb.barrier.CopyFrom(pb_barrier)
 
+    @dispatch(FFConfigurator)
+    def _serialize(self, ffconfigurator: FFConfigurator):
+        pb_ffc = pb.FFConfigurator()
+        pb_ffc.name = ffconfigurator.name
+        pb_ffc.offset = ffconfigurator._offset
+        pb_ffc.block_circuit_size = ffconfigurator._blocked_circuit_size
```
Aubaert, line 147: I'm not sure we want to serialize this, this is just a boolean that indicates if the Configurator has been placed in an experiment.
If we serialize and deserialize it on its own, we lose the reference of the deserialized object in said experiment, so we don't need to block the size. If we serialize it in an experiment and we deserialize the experiment, this boolean will be set to True when we add the deserialized Configurator to the deserializing Experiment.
The only case where it could be True without an Experiment is if a user calls this method themselves, but I assume this is never the case

perceval/serialization/_circuit_serialization.py
```diff
@@ -138,6 +139,51 @@ def _serialize(self, barrier: comp.Barrier):
...
+            pb_ffc.configs[str(state)].CopyFrom(pb_vars)
+        self._pb.ff_configurator.CopyFrom(pb_ffc)
+
+    @dispatch(FFCircuitProvider)
+    def _serialize(self, ffcp: FFCircuitProvider):
+        from ._experiment_serialization import serialize_experiment
+        pb_ffcp = pb.FFCircuitProvider()
+        pb_ffcp.name = ffcp.name
+        pb_ffcp.offset = ffcp._offset
+        pb_ffcp.block_circuit_size = ffcp._blocked_circuit_size
```
Aubaert, line 170: Same here

perceval/serialization/_component_deserialization.py
```diff
@@ -97,5 +99,37 @@ def deserialize_pbs(_, __) -> comp.PBS:
...
 
 
-def deserialize_barrier(m: int, serial_barrier) -> comp.Barrier:
+def deserialize_barrier(m: int, serial_barrier, _) -> comp.Barrier:
     return comp.Barrier(m, serial_barrier.visible)
+
+
+def deserialize_ff_configurator(m: int, serial_ffc, known_params: dict = None) -> FFConfigurator:
+    from .deserialize import deserialize_circuit
+    default_config = {k: v for k, v in serial_ffc.default_config.mapping.items()}
```
Aubaert, line 108: Does `default_config = dict(serial_ffc.default_config.mapping)` work ?

perceval/serialization/deserialize.py
```diff
@@ -365,20 +371,32 @@ def resolve(self):
...
             func = self.deserialize_detector_fn[t]
             detector = func(serial_sub_comp)
-            experiment.add(i, detector)
+            detectors[i] = detector
+
+        # Needed before adding the heralds
+        for serial_comp in self._pb_e.components:
+            component = CircuitBuilder.deserialize(serial_comp, self._params)
+            if isinstance(component, AFFConfigurator):
+                for i in range(serial_comp.starting_mode, serial_comp.starting_mode + component.m):
```
Aubaert, line 388: Why can't we use `experiment.add` in this case?

perceval/serialization/deserialize.py
```diff
@@ -365,20 +371,32 @@ def resolve(self):
...
+                    if not injected_detectors[i]:
+                        experiment._components.append(((i,), detectors[i]))
+                        injected_detectors[i] = True
+                experiment._components.append((tuple(i for i in range(serial_comp.starting_mode, serial_comp.starting_mode + component.m)), component))
+            else:
+                experiment.add(serial_comp.starting_mode, component)
+
+        # Add detectors
+        for i, d in enumerate(detectors):
+            if d is not None:
```
Aubaert, line 398: Shouldn't we check that `injected_detectors[i] == False` ? I don't see what prevents adding twice the same detector

tests/test_serialization.py
```diff
@@ -317,6 +277,21 @@ def test_experiment_serialization():
...
+
+    mzi = catalog['mzi phase last'].build_circuit()
+    ffc = FFConfigurator(2, 0, mzi, {"phi_a": 0, "phi_b": 0}, "ffc name")
+    ffc.add_configuration((0, 1), {"phi_a": 3, "phi_b": 0})
+    ffc.add_configuration((1, 0), {"phi_a": 0, "phi_b": 3})
+    e.add(0, ffc)
+
+    ffcp = FFCircuitProvider(1, 0, Circuit(2) // BS.H(), "ffcp name")
+    ffcp.add_configuration((0,), Circuit(2) // BS.Rx())
+    subexp = Experiment(2)
```
Aubaert, line 289: Does it work if we try to add a `Processor` here as it is the case in pereval 0.12 ?
My take is that the `Configurator` should extract the experiment from the `Processor`

tests/imgs/test_svg_dump_feed_forward_phys.svg
AliseeNguyen, line 1: I cannot see what changed in the new images

tests/_test_utils.py
```diff
@@ -37,12 +37,15 @@
 from unittest.mock import MagicMock
 
 import perceval as pcvl
+from perceval.components import ACircuit, PS, IDetector, AFFConfigurator, FFConfigurator, FFCircuitProvider, \
+    BSLayeredPPNR, Detector, Experiment, Circuit
```
AliseeNguyen, line 41: Since you import many things here, maybe you can remove the previous import of perceval and just add the imports of AProcessor and pdisplay_to_file

AliseeNguyen (approved):

tests/imgs/test_svg_dump_feed_forward_phys.svg
Aubaert: When we test the SVGs, we don't look at the rendering but at what is inside the SVG. Since the order of components changes the inside of the SVG, we need to change the target even if it has the same rendering

perceval/serialization/_circuit_serialization.py
ericbrts: As `block_circuit_size()` is a public method from the `FFConfigurator` class, I was expecting to have to serialize it, as we can't deduce this boolean value from the context where it's being used.

perceval/serialization/deserialize.py
ericbrts: Because `experiment.add` adds `Barriers` and injects `Detectors` inside the component list automatically. As these are already serialized, I don't want them to be re-added automatically on top of the deserialized ones.

perceval/serialization/deserialize.py
ericbrts: In an `Experiment`, `Detector` instances appear twice when they're on a feed-forwarded mode. We definitely want to add all the detectors to the detector list because it's this list that's used during the output states post-processing.

tests/imgs/test_svg_dump_feed_forward_phys.svg
ericbrts: Yes, sorry that was the point : nothing changed visually - but the test failed because of the SVG structure.

raksharuia (approved):

Aubaert (approved):

PR #585, opened 2025-04-09
Pcvl 936 experiments add port should behave differently when we add an herald

Aubaert:


perceval/components/_mode_connector.py
```diff
@@ -68,11 +68,11 @@ def __init__(self, left_processor, right_obj, mapping):
             left_processor = left_processor.experiment
         self._lp = left_processor
         self._ro = right_obj  # Can either be a component or a processor
-        self._r_is_component = isinstance(right_obj, AComponent)  # False means it is a Processor
+        self._r_is_component = isinstance(right_obj, AComponent)  # False means it is an Experiment
```
raksharuia, line 71: Not really a part of change in this ticket, however -> the first input argument is called `left_processor` but the class documentation says that it should be an `Experiment`. It would be nice to update the argument name for clarity. Typehinting is also missing here.

Also, the documentation does not mention that the `right_obj` can be an experiment, it simply says a component or a processor but the if the check in L71 is False, comment says it must be an experiment. 

Also, should not the ` right_obj` be plugged to the **right** of the `left_obj` ?

perceval/components/abstract_processor.py
```diff
@@ -259,6 +260,10 @@ def m(self) -> int:
         """
         return self.experiment.m
 
+    @property
+    def m_in(self) -> int:
```
raksharuia, line 264: Missing documentation.

perceval/serialization/deserialize.py
```diff
@@ -380,8 +378,8 @@ def resolve(self):
             detector = func(serial_sub_comp)
             experiment.add(i, detector)
 
-        self.deserialize_ports(experiment, self._pb_e.input_ports, PortLocation.INPUT, True)
-        self.deserialize_ports(experiment, self._pb_e.output_ports, PortLocation.OUTPUT, False)
+        self.deserialize_ports(experiment, self._pb_e.input_ports, PortLocation.OUTPUT, True)
+        self.deserialize_ports(experiment, self._pb_e.output_ports, PortLocation.INPUT, False)
```
raksharuia, line 382: This seems a bit confusing to me `self._pb_e.output_ports` are with `PortLocation.INPUT` and vice versa in the previous line. I do not think I am following correctly - could you please clarify what is happening here?

perceval/components/abstract_processor.py
```diff
@@ -237,15 +237,16 @@ def add(self, mode_mapping, component, keep_port: bool = True) -> AProcessor:
...
 
-    def add_herald(self, mode: int, expected: int, name: str = None) -> AProcessor:
+    def add_herald(self, mode: int, expected: int, name: str = None, location: PortLocation = PortLocation.IN_OUT) -> AProcessor:
         r"""
         Add a heralded mode
 
         :param mode: Mode index of the herald
         :param expected: number of expected photon as input AND output on the given mode (must be 0 or 1)
         :param name: Herald port name. If none is passed, the name is auto-generated
+        :param location: Port location of the herald (input, output or both)
```
raksharuia, line 247: Line 245 documents `expected` photons to be 1 or 0 but there isn't any check to prevent a user from setting a higher value for heralded inputs/outputs.

perceval/components/_mode_connector.py
Aubaert: For the type hinting, it's not possible to put it here since experiment.py imports this file (this would result in circular import).

I don't understand your last point.

perceval/components/experiment.py
```diff
@@ -670,17 +722,17 @@ def detection_type(self) -> DetectionType:
         return get_detection_type(self._detectors)
 
     @property
-    def heralds(self):
-        pos = {}
-        for port, port_range in self._out_ports.items():
-            if isinstance(port, Herald):
-                pos[port_range[0]] = port.expected
-        return pos
+    def heralds(self) -> dict[int, int]:
```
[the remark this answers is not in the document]
Aubaert: The idea of keeping `heralds` for the output was to simplify the conversion as we use the output heralds most of the times (same for `m`), but we can discuss about depreciating it and use `output_heralds` as a preferred call

perceval/serialization/deserialize.py
Aubaert: It was a mistake (my bad)

AliseeNguyen (approved):

PR #594, opened 2025-05-19
PCVL-960 automatic state generator

Aubaert:


perceval/utils/stategenerator.py
```diff
@@ -193,3 +194,54 @@ def dicke_state(self, n: int, k: int = None) -> StateVector:
...
+        return state
+
+    @staticmethod
+    def evenly_spaced_state(n: int, m: int) -> BasicState:
+        """
+        Get a state where photons are evenly spaced. Ignores the encoding.
+
+        :param n: Number of photons
+        :param m: Number of modes.
+        :return: BasicState where the photons are the most spread as possible in the modes.
```
raksharuia, line 235: either `as spread out as possible`or `the most spread out possible`

tests/test_stategenerator.py
```diff
@@ -215,3 +215,39 @@ def test_dicke_state(mock_warn, encoding):
...
+
+def test_zero_padded_state():
+    n = 2
+    m = 5
+
+    state = pcvl.StateGenerator.zero_padded_state(n, m)
+
+    assert state.n == n
+    assert state.m == m
+    assert state[:n] == pcvl.BasicState(n * [1])
```
raksharuia, line 228: maybe we should also assert state[n:] are 0 filled. It is kind of stupid though, so not really necessary.

raksharuia (approved):

tests/test_stategenerator.py
raksharuia: Yup, it is redundant indeed. My idea was - if by any other error, we ended up adding some garbage instead of 0s, it would be caught. But that would technically mean something else is really wrong with the code :-P. So let it be.

raksharuia (approved):

PR #596, opened 2025-05-21
PCVL 967 Experiment are deep copied

Benoit-F-Q:


perceval/utils/postselect.py
```diff
@@ -39,7 +39,8 @@
     from typing_extensions import TypeAlias  # Only used with python 3.9
 
 PostSelect: TypeAlias = xq.PostSelect
-
+# Temporary implementation of __deepcopy__ until exqalibur release
```
Aubaert, line 42: Can you add a ticket for this?

Aubaert (approved):

PR #597, opened 2025-05-26
PCVL-964 add build_experiment() to CatalogItem's interface

Benoit-F-Q:
I kept the `build_processor()` in `CatalogItem`'s interface, but its implementation is always a simple `_init_processor()` call.
Should `build_processor()` stop being an abstract method?


Aubaert (approved):
About your question, I would say yes, as I don't see any reason why the `build_processor` could give a different result than `self._init_processor`

PR #601, opened 2025-06-04
PCVL-925 Fix PersistentData.is_writable

AliseeNguyen:


raksharuia (approved):
The test passes for me too. I don't have any other suggestions.

PR #602, opened 2025-06-04
PCVL-977 rename JobGroup.list_existing() into JobGroup.list_locally_s…

Benoit-F-Q:
…aved()


Aubaert (commented):
Can you add this change in the legacy.rst file?

Aubaert (approved):

PR #606, opened 2025-06-06
PCVL-973 add 'cloud maximal job count' to RemoteConfig

Benoit-F-Q:


perceval/runtime/job_group.py
```diff
@@ -408,7 +420,10 @@ def run_sequential(self, delay: float) -> None:
 
         :param delay: number of seconds to wait between launching jobs on cloud
         """
-        self._launch_jobs(rerun=False, delay=delay)
+        self._launch_jobs(await_responses = True,
+                          rerun=False,
+                          nb_parallel = RemoteConfig.get_cloud_maximal_job_count() or 1,
```
Aubaert, line 425: I think this should always be 1 for run_sequential, and should use the maximal job count only for run parallel.

Aubaert (approved):

perceval/runtime/job_group.py
```diff
@@ -350,56 +351,67 @@ def list_unsent_jobs(self) -> list[RemoteJob]:
...
-            if rerun and job.is_failed:
-                job = job.rerun()
-                if replace_failed_jobs:
-                    self._jobs[job_idx] = job
+        progress_bar = None
+        if await_responses:
+            bar_format = '{percentage:3.0f}%|{bar}|{n_fmt}/{total_fmt}|{desc}'
+            progress_bar = tqdm(total=len(jobs_to_run), bar_format=bar_format, desc="Successful: 0, Failed: 0")
+
+        maximal_concurent_jobs = nb_parallel if nb_parallel else len(jobs_to_run)
```
AliseeNguyen, line 373: can be simplified in : `maximal_concurrent_jobs = nb_parallel or len(jobs_to_run)`

perceval/runtime/job_group.py
```diff
@@ -350,56 +351,67 @@ def list_unsent_jobs(self) -> list[RemoteJob]:
...
+        peek_delay = 1 if await_responses else 0
```
AliseeNguyen, line 374: and this in : `peek_delay = int(await_responses)`

AliseeNguyen (approved):

perceval/runtime/job_group.py
Benoit-F-Q: I think there is no point launching jobs 1 by 1 if you have more compute slots available

I think there is little use launching only some (# = cloud_maximal_job_count) jobs and leave if you have more ready, because we provide no way to track ongoing jobs.

I think the main use case is launching all prepared jobs, with maximal concurrency (the script is thus blocking and deserves its progress_bar), but neither `run_sequential` nor `run_parallel` is designed to

perceval/runtime/job_group.py
Benoit-F-Q: pas d'accord ici, 1 est une valeur explicite qui pourra être ajustée en 2.8 ou 0.25

perceval/runtime/job_group.py
raksharuia: What if a user had permissions to send  only 1 job on cloud? They would want to launch them sequentially 1 by 1.

Benoit-F-Q:
- run_parallel is always blocking, with progress bar, even if only 1 job (we should consider a ticket for a non-blocking version using a background thread)

- Added a warning if `RemoteConfig.set_cloud_maximal_job_count()` is not set

- Increaded 'peek delay' to 10s

PR #613, opened 2025-06-24
Fix AComponent.get_parameters and Expressions.

anthonyrtw:
- Corrected behaviour of the `Expression.fixed` property: an expression is now considered fixed if all sub-parameters are fixed.
- Fix `AComponent.get_parameters` to return expected results when `expressions=True`.

## Bugs

- Previously, `Expression` inherited its `fixed` property from its parent class `Parameter` which determines if a parameter is fixed by checking if it has a sympy symbol. This logic did not work for Expressions.

- `AComponent.get_parameters(expressions=True)` did not return expressions in certain instances due to the `_params` attribute behaving differently for components vs. circuits.

## Fixes 

- A new dedicated property for `Expression` has been defined. This checks if all its base parameters are defined.

- A new hidden attribute `_raw_params` is defined for abstract components, `AComponent`. This contains the "highest-level" parameters (incl. hidden parameters) as they appear in the circuit.  
```python
A, B = P("A"), P("B")
circuit = Circuit(2) // PS(A) // BS(2 * B)
circuit._raw_params
```
```
{'A': Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586),
 'max_error': Parameter(name='max_error', value=0, min_v=0.0, max_v=3.141592653589793),
 '(2*B)': Expression('2*B', parameters={Parameter(name='B', value=None)}),
 'phi_tl': Parameter(name='phi_tl', value=0, min_v=0.0, max_v=6.283185307179586),
 'phi_bl': Parameter(name='phi_bl', value=0, min_v=0.0, max_v=6.283185307179586),
 'phi_tr': Parameter(name='phi_tr', value=0, min_v=0.0, max_v=6.283185307179586),
 'phi_br': Parameter(name='phi_br', value=0, min_v=0.0, max_v=6.283185307179586)}
```

`AComponent.get_parameters` then cycles through `_raw_params` and returns all lowest-level parameters for `expressions=False` and all highest level parameters for `expressions=True`.

## Examples 

For a circuit:
```python 
from perceval import Circuit, PS, BS, P

A, B = P("A"), P("B")
circuit = Circuit(2) // PS(A) // BS(A, phi_tl=2*A, phi_bl=B)
print(circuit.get_parameters(), '\n')
print(circuit.get_parameters(all_params=True), '\n')
print(circuit.get_parameters(expressions=True), '\n')
print(circuit.get_parameters(all_params=True, expressions=True), '\n')
```

For a component:
```python 
A, B = P("A"), P("B")
bs = BS(A, phi_tl=2*A, phi_bl=2*B)
print(bs.get_parameters(), '\n')
print(bs.get_parameters(all_params=True), '\n')
print(bs.get_parameters(expressions=True), '\n')
print(bs.get_parameters(all_params=True, expressions=True), '\n')
```

Both of these cells output:
```
[Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586), Parameter(name='B', value=None, min_v=0.0, max_v=6.283185307179586)] 

[Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586), Parameter(name='max_error', value=0, min_v=0.0, max_v=3.141592653589793), Parameter(name='B', value=None, min_v=0.0, max_v=6.283185307179586), Parameter(name='phi_tr', value=0, min_v=0.0, max_v=6.283185307179586), Parameter(name='phi_br', value=0, min_v=0.0, max_v=6.283185307179586)] 

[Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586), Expression('2*A', parameters={Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586)}), Parameter(name='B', value=None, min_v=0.0, max_v=6.283185307179586)] 

[Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586), Parameter(name='max_error', value=0, min_v=0.0, max_v=3.141592653589793), Expression('2*A', parameters={Parameter(name='A', value=None, min_v=0.0, max_v=6.283185307179586)}), Parameter(name='B', value=None, min_v=0.0, max_v=6.283185307179586), Parameter(name='phi_tr', value=0, min_v=0.0, max_v=6.283185307179586), Parameter(name='phi_br', value=0, min_v=0.0, max_v=6.283185307179586)]
```

## Extra Notes

These changes support the upcoming addition of parameter expression handling in [MerLin](https://github.com/merlinquantum/merlin).


perceval/components/abstract_component.py
```diff
@@ -113,16 +114,17 @@ def get_parameters(self, all_params: bool = False, expressions = False) -> list[
             If False, returns the raw parameters that make up the expressions only. Default `False`.
         """
         param_list = []
-        for param in self._params.values():
+        for param in self._raw_params.values():
```
Aubaert, line 117: This method had already the expected behaviour for components. 
Instead of changing its behaviour to fix the behaviour of circuit, which requires adding your `_raw_params` member that can quickly lead to inconsistencies, it would be better to redefine this method in ACircuit that would call this method for each subcomponent and merge the results (just beware that all components don't have this method)

perceval/components/abstract_component.py
```diff
@@ -113,16 +114,17 @@ def get_parameters(self, all_params: bool = False, expressions = False) -> list[
...
+                        # Add raw expression
                         if param not in param_list:
-                            param_list.append(param)
+                            param_list.append(param) 
                     else:
-                        for p in param.parameters:
-                            if p not in param_list:
-                                param_list.append(p)
+                        # Add subparameters within expression
+                        param_list = list(set(param_list) | param._params)
```
Aubaert, line 126: Beware that using `set` can lead to random ordering of the results (though I agree that this is faster), whereas this method currently gives the parameters in the order they were added.

Aubaert (changes requested):

ericbrts (approved):

PR #622, opened 2025-07-10
PCVL-991 Modes displayed on top of the port

Aubaert:


AliseeNguyen (approved):
We will need to relaunch Qiskit and MyQLM notebooks in interop

PR #624, opened 2025-07-15
PCVL-997 Generate payload without connecting

AliseeNguyen:


perceval/runtime/payload_generator.py
```diff
@@ -0,0 +1,97 @@
...
+        if platform_name:
+            j['platform_name'] = platform_name
+
+        payload = {
+            'command': command,
+            **kwargs
+        }
+
+        if experiment.input_state:
+            experiment.check_input(experiment.remove_heralded_modes(experiment.input_state))
```
Aubaert, line 89: `check_input` is redefined by RemoteProcessor. The check using `experiment` has already been done when using `with_input`. 
The best is to put it in `RemoteProcessor.prepare_job_payload`

Aubaert (changes requested):

Aubaert (approved):

PR #642, opened 2025-08-21
PCVL-896-adapt-and-transmit-the-heralds-and-post-select-between-simulation-layers

Aubaert:


perceval/simulators/delay_simulator.py
```diff
@@ -124,6 +125,16 @@ def _postprocess_sv_impl(self, results: StateVector) -> StateVector:
                 output += probampli * reduced_out_state
         return output
 
+    def _transmit_heralds_postselect(self):
+        heralds = None
```
AliseeNguyen, line 129: I don't understand how this code can be reached since _can_transmit_selection is False (from ASimulatorDecorator)

Aubaert: True, fixed

AliseeNguyen (commented):
Is it possible to add some tests ? I saw only one for DelaySimulator and postselect

(and this test is now failing after your fix)

perceval/simulators/delay_simulator.py
```diff
@@ -124,6 +128,16 @@ def _postprocess_sv_impl(self, results: StateVector) -> StateVector:
                 output += probampli * reduced_out_state
         return output
 
+    def _transmit_heralds_postselect(self):
+        heralds = self._heralds
```
AliseeNguyen, line 132: maybe initialize heralds with an empty dict since it will be changed afterward

perceval/simulators/delay_simulator.py
```diff
@@ -124,6 +128,16 @@ def _postprocess_sv_impl(self, results: StateVector) -> StateVector:
...
+        if self._heralds:
+            heralds = {m + self._mode_range()[0]: v for m, v in self._heralds.items()}
+        postselect = self._postselect
```
AliseeNguyen, line 135: same

AliseeNguyen (approved):

PR #650, opened 2025-08-29
PCVL-1048 simulation with noisy fs

Aubaert:


perceval/components/experiment.py
```diff
@@ -810,6 +809,10 @@ def with_input(self, input_state: FockState) -> None:
         self._input_state = FockState(input_list)
         self._input_changed()
 
+    @dispatch(NoisyFockState)
```
ericbrts, line 812: Could be factored with the `@dispatch(StateVector)` version of with_input

PR #653, opened 2025-09-01
PCVL-1058 SLAP wrong mask init

Aubaert:


perceval/backends/_slap.py
```diff
@@ -47,8 +47,8 @@ def set_circuit(self, circuit: ACircuit):
         super().set_circuit(circuit)  # Computes circuit unitary as _umat
         self._slap.set_unitary(self._umat)
 
-    def set_mask(self, masks: str | list[str], n = None):
-        super().set_mask(masks, n)
+    def _init_mask(self):
```
AliseeNguyen, line 50: I don't see why it solves the problem, as `self._mask` is only valued in `_init_mask` with the check `self._input_state is not None`, and `set_mask` calls `init_mask`

Aubaert: `init_mask` is also called at the end of `set_input_state`

AliseeNguyen (approved):

AliseeNguyen (approved):
I did some tests, ok for me

PR #661, opened 2025-09-04
added ref to training center

samuelquandela:
I want to make a bridge between the docs and the training centre.
For now I just added the link but later I also want to link specific lessons.

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -88,3 +88,7 @@ Probabilities: {
 }
 
 Now that you can run some code, let's continue with a tutorial to learn Perceval syntax.
+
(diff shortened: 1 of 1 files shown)


docs/source/getting_started.rst
```diff
@@ -88,3 +88,7 @@ Probabilities: {
 }
 
 Now that you can run some code, let's continue with a tutorial to learn Perceval syntax.
+
+You can also check out the Quandela Training Centre_ for more Lessons and Tutorials.
```
Aubaert, line 92: Hello, I'm not sure this is the right place to put this link.
I think it would be more suited in the index.rst file (perhaps in the `related projects` or after the link to `Welcoming Contributors`)

PR #667, opened 2025-09-19
PCVL-1071 expression copy in processor

Aubaert:


perceval/components/experiment.py
```diff
@@ -234,7 +235,11 @@ def copy(self, subs: dict | list = None) -> Experiment:
         Performs a deep copy of the current experiment.
         """
         get_logger().debug(f"Copy experiment {self.name}", channel.general)
+        if subs is None:
+            subs = {}
+        Experiment._no_copiable_attributes.add("_components")
```
ericbrts, line 240: Why is `_components` not permanently in the non copiable attributes list?

ericbrts (approved):

perceval/components/experiment.py
Aubaert: It's just that since we replace it by an empty list on the next line, it's not necessary to copy it in the deepcopy

ericbrts: Ok its purpose is to optimise this code a bit. But `_components` still has to be copiable elsewhere (?)

PR #717, opened 2026-02-27
Allow Experiment input states to be lists/tuples.

anthonyrtw:
The vast majority of the time, users define input states as lists when defining Experiments. However, the list/tuple must be converted into a `BasicState` before passing into `Experiment.with_input`. e.g.

```python
experiment.with_input(BasicState([1, 2, 0, 1, 2]))
```

For conciseness, and to lower the need for excessive class instantiation on the user's part, I have introduced a new overloaded method for `Experiment.with_input` that accepts lists & tuples.

```python
experiment.with_input([1, 2, 0, 1, 2])
```

To avoid possible confusion between logical states and Fock states when using lists/tuples, I have created an `AssertionError` so that list/tuple type inputs cannot work with processors which have been encoded with ports.

```
AssertionError: Input state of type `tuple` on an experiment with ports is ambiguous. Please use input of type `LogicalState`.
```

### DIFF
--- a/perceval/components/experiment.py
+++ b/perceval/components/experiment.py
@@ -36,7 +36,7 @@ from typing import Callable
 from multipledispatch import dispatch
 
 from perceval.utils import FockState, AnnotatedFockState, Parameter, PostSelect, LogicalState, NoiseModel, ModeType, StateVector, \
-    SVDistribution, NoisyFockState
+    SVDistribution, NoisyFockState, BasicState
 from perceval.utils.logging import get_logger, channel
 from perceval.utils.algorithms.simplification import perm_compose, simplify
 from ._mode_connector import ModeConnector, UnavailableModeException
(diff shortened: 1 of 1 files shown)


Benoit-F-Q:
"explicit is better than implicit"
We expect the user to tell what he's doing. The cost of typing `FockState(` and ')' is considered small.

Benoit-F-Q:
> In my opinion, "BasicState" is not explicit coding. To be honest, I don't really understand the concept of "BasicState" and it's more or less the same than "FockState".
It is explicit in the developper's sense : you tell precisely what you intend, instead of giving an array which will be (implicitely) converted to what you meant.

A BasicState is basic in the sense that it is not a superposition (StateVector).
It can be a FockState (no annotation, best efficiency), a NoisyFockState (photons labeled with integers) or an AnnotatedFockState (most versatile, much heavier).

I don't know where the name BasicState did come from, FockState would have been more consistent. But we're quite stuck with it.

> I think that if users do `experiment.with_input([1, 2, 0, 1, 2])` they will know that they are inputting a certain number of photons in certain mode.

Except when they are not, as `experiment.with_input([1, 2, 0, 1, 2])` could mean a LogicalStates, as Anthony mentionned.

anthonyrtw:
> To avoid possible confusion between logical states and Fock states when using lists/tuples, I have created an AssertionError so that list/tuple type inputs cannot work with processors which have been encoded with ports.

Maybe my exception raises more issues than solves. It would probably be better to explicitly **always** interpret a `list` as a basic state.
Experiments with encoding ports can raise an error if the `list` input state length does not match the number of modes.
```python
if self._in_ports:
    assert len(input_state) == self.m, f"Input state length does not match processor size. Expected length {self.m}, got {len(input_state)}. For `list` inputs, please ensure that your input represents the full Fock state."
```

Aubaert:
Hello, the modifications and commit proposed by this PR have been merged in another PR, so this PR is now a duplicate.

PR #725, opened 2026-03-18
Allow Experiment.set_postselection to accept string.

anthonyrtw:
Post-selection rules are defined using strings which contain logical instructions on how to postprocess the results.
The `PostSelect` class verifies & processes the logic of these strings.

A common friction point for a user is setting the postselection in an Experiment. Here, it is common for the user to pass a string to `Experiment.set_postselection`. Even when users know that a `PostSelect` object is required, passing a raw string is often the most natural thing to do when coding on the fly. Since the string already encodes the full logic, this leads to an avoidable `TypeError`.

```python
experiment.set_postselection("[0, 1, 2] == 1 AND [3, 4] < 2 OR [5] == 1") # Error
  
experiment.set_postselection(PostSelect("[0, 1, 2] == 1 AND [3, 4] < 2 OR [5] == 1"))
```

To prevent this error, input `str` arguments in `Experiment.set_postselection` are wrapped as `PostSelect` objects before proceeding.

Also, I have noticed a bug when a user attempts to do: `experiment.set_postselection(None)` as mentioned in the function docstring. Here, this would formerly raise a `TypeError`. I have amended the function to allow `None` arguments to clear the post-selection rule.


perceval/simulators/noisy_sampling_simulator.py
```diff
@@ -176,7 +176,7 @@ def compute_physical_logical_perf(self, value: bool):
 
     def set_selection(self,
                       min_detected_photons_filter: int = None,
-                      postselect: PostSelect = None,
+                      postselect: PostSelect = PostSelect(),
```
Aubaert: Be careful with non-immutable default arguments, as they are kept alive for the whole script duration (so if someone merges a PostSelect into this one, it will affect all default PostSelect)

perceval/simulators/simulator_interface.py
```diff
@@ -81,7 +81,7 @@ def set_heralds(self, heralds: dict[int, int]):
 
     def set_selection(self,
                       min_detected_photons_filter: int = None,
-                      postselect: PostSelect = None,
+                      postselect: PostSelect = PostSelect(),
```
Aubaert: Same

Aubaert (approved):

PR #732, opened 2026-05-06
PCVL-1214 protobuf file generated in exqalibur package

Benoit-F-Q:


Aubaert:
If we change the workflow, can you adapt the pipelines so the tests are launched when merging on develop?

Aubaert (approved):

PR #743, opened 2026-05-25
PCVL-1245 Added flaky to test + increased relative acceptability

Aubaert:


tests/simulators/test_simulator.py
```diff
@@ -438,6 +440,7 @@ def test_evolve_density_matrix():
     assert max((final_dm.mat-comparing_dm.mat).data) < 1e-10
 
 
+@flaky(max_runs=2)
```
Benoit-F-Q: max_runs=3 ?

Benoit-F-Q (approved):

tests/simulators/test_simulator.py
Aubaert: OK

PR #745, opened 2026-05-25
PCVL-1147 support for coherent light inputs

Aubaert:


jeonghoon-ad:
`CoherentState.__ne__` currently recurses:

```python
def __ne__(self, other: CoherentState) -> bool:
    return not self != other
```

So any `!=` comparison raises `RecursionError`:

```python
CoherentState([1]) != CoherentState([2])
```

Local check on this PR branch:

```text
eq_same True
eq_diff False
ne_diff_error RecursionError maximum recursion depth exceeded
ne_same_error RecursionError maximum recursion depth exceeded
```

A direct fix is:

```python
def __ne__(self, other: CoherentState) -> bool:
    return not self == other
```

A regression test can cover both equal and non-equal states:

```python
assert not (CoherentState([1]) != CoherentState([1]))
assert CoherentState([1]) != CoherentState([2])
```

PR #752, opened 2026-06-10
PCVL-1261 Added methods to apply phase noise with a chosen rng

Aubaert:


perceval/components/abstract_component.py
```diff
@@ -63,6 +64,9 @@ def is_composite(self) -> bool:
         """
         return False
 
+    def apply_phase_noise(self, phase_error = 0, phase_imprecision = 0, rng: random.Random = None):
+        """Applies a noise on phases. Acts in-place"""
```
Benoit-F-Q, line 68: Seems it's missing a minimal body

perceval/components/feed_forward_configurator.py
```diff
@@ -256,3 +262,28 @@ def configure(self, measured_state: BasicState) -> ACircuit:
 
     def circuit_template(self) -> ACircuit:
         return self._controlled
+
+    @staticmethod
+    def _apply_phase_noise_on_one_config(config: dict[str, float],
+                                         phase_error: float,
+                                         phase_imprecision: float,
+                                         rng: random.Random):
+        # TODO: We have a problem here: the parameters could be anything and not just phases
```
Benoit-F-Q: Filter on parameter name?

Anyway, the phase error should be applied to components, not to parameters. Delay the noise application until we create thr circuit ?

perceval/components/feed_forward_configurator.py
```diff
@@ -256,3 +262,28 @@ def configure(self, measured_state: BasicState) -> ACircuit:
...
+            if phase_imprecision > 0:
+                val = phase_imprecision * round(val / phase_imprecision)
+
+            config[name] = val
+
+    def apply_phase_noise(self, phase_error = 0, phase_imprecision = 0, rng: random.Random = None):
+        if rng is None:
+            rng = random.Random()
+
+        self._apply_phase_noise_on_one_config(self._default_config, phase_error, phase_imprecision, rng)
```
Benoit-F-Q: Note: `_default` config was unused until now. Wasn't something missing?

perceval/components/abstract_component.py
Aubaert: We don't need to write "pass" when there is a docstring, but I'll add it for clarity

perceval/components/feed_forward_configurator.py
Aubaert: The default config is only used to assign the parameters to the default circuit in the init, and to serialize/deserialize the object

Benoit-F-Q (approved):

PR #753, opened 2026-06-10
CW-202 use emt workflow remote worker

Aubaert:


perceval/runtime/payload_generator.py
```diff
@@ -131,3 +132,17 @@ def from_computation(computation: Computation,
         if noise is not None:
             payload[KEY_NOISE] = noise
         return payload
+
+    @staticmethod
+    def configure_computer_from_payload(computer: AbstractComputer, payload: dict) -> Computation:
```
Benoit-F-Q: Do we agree on its purpose? Changing Computer's configuration should be a rare event

perceval/serialization/_serialization_registrator.py
```diff
@@ -0,0 +1,116 @@
...
+        - "deserialize": Callable[[str], cls] static method, that is used to produce a class instance from the result of "serialize".
+
+    :param cls: The class to register in serialization.
+    :param tag: The tag to use as an identifier for this class. Defaults to the class name.
+    :param default_compress: Whether to compress the resulting string by default.
+    """
+    if tag is None:
+        tag = cls.__name__
+
+    if hasattr(cls, "serialize"):  # Use a more reserved name (perceval_serialize ?)
```
Benoit-F-Q: register class + [de]serialize free functions instead ?

perceval/runtime/payload_generator.py
Aubaert: The idea is to use it every time we receive a payload (like in the RemoteComputer tests). Since we no longer have a payload inside the execute(), we won't use it inside it

perceval/runtime/abstract_computer.py
```diff
@@ -249,3 +338,30 @@ def performance(self):
     @abstractmethod
     def type(self):
         pass
+
+    def apply_configuration(self,
+                            mitigations: list[AbstractMitigation] = None,
+                            noise: NoiseModel = None,
+                            parameters: dict[str, Any] = None):
```
Benoit-F-Q: "-> ContextManager" ?

perceval/runtime/abstract_computer.py
```diff
@@ -249,3 +338,30 @@ def performance(self):
...
+        """
+        :param mitigations: The mitigations to apply within the ContextManager. If None, nothing is changed
+        :param noise: The noise model to apply within the ContextManager. If None, nothing is changed
+        :param parameters: The parameters to apply within the ContextManager. If None, nothing is changed
+        :return: A ContextManager that applies the given arguments to the computer (noise, mitigations, parameters)
+          at enter and reset the parameters to the previous values at exit
+        """
+        # TODO: For RemoteComputer, we need to overload? Or use a public way to get the mitigations?
```
Benoit-F-Q: I don't get it

perceval/runtime/abstract_computer.py
```diff
@@ -249,3 +338,30 @@ def performance(self):
...
+        starting_mitigations = self._error_mitigations if mitigations is not None else None
+        starting_noise = self.noise if noise is not None else None
```
Benoit-F-Q, line 359: Pitfall: if self.noise is None, it can't be reverted

Benoit-F-Q (approved):

PR #763, opened 2026-06-26
Allow string arguments for interferometer shape in GenericInterferometer

anthonyrtw:
`GenericInterferometer` currently only supports the `InterferometerShape` enum as argument for `shape`. Other parts of Perceval such as in `Circuit.decomposition` allow "triangle" and "rectangle" strings as shape arguments.

For package-wide consistency and ease of typing, I use the same logic in `Circuit.decomposition` for automatically promoting user string arguments to `InterferometerShape`.

```python
# New available UI
gi = GenericInterferometer(m, fun_gen=fun_gen, shape="triangle")

# Still supported
gi = GenericInterferometer(m, fun_gen=fun_gen, shape=InterferometerShape.TRIANGLE)
```

Under the hood:
```python
        if isinstance(shape, str):
            try:
                shape = InterferometerShape[shape.upper()]
            except:
                raise ValueError(f"Unknown interferometer shape: {shape}")
```

Side note:
I am a known Python enum hater.


Benoit-F-Q:
Unless you use a plain text editor, strings are far more error-prone, because `_validate_shape` is delayed to execution.
Any modern python IDE - even emacs or vim with a python setup - will check the validity of the enum.

This helps both when you call `GenericInterferometer(m, fun_gen=fun_gen, shape="trinagle")` and when you decide to replace `triangle` by `linear` (but only for GenericInterferometer, not for Circuit.decomposition

Benoit-F-Q (approved):

PR #765, opened 2026-06-29
Loss balancing

anthonyrtw:
Add Loss Balancing class to Perceval.

Outstanding work: Determine how calibration data necessary for mitigation is fed into `_parse_results`


Aubaert:
Added in https://github.com/Quandela/Perceval/pull/772

PR #766, opened 2026-06-29
Add Kipu Quantum Hub provider (perceval.providers.kipu)

wagnerse-kipu:
Adds an optional cloud provider, `perceval.providers.kipu`, that submits jobs through the
[Kipu Quantum Hub](https://dashboard.hub.kipu-quantum.com/). The Hub routes quantum jobs to multiple hardware providers. Through it, Perceval can run the Quandela photonic backends exposed
on the Hub — `quandela.sim.belenos` and `quandela.qpu.belenos`, which are the Hub's ids for
Quandela's native `sim:belenos` and `qpu:belenos` platforms.

**Adds**
- `perceval/providers/kipu/` — `Session` + `KipuRPCHandler`, same pattern as the `quandela`/`scaleway` providers.
- Auth via Kipu Personal Access Token (or `qhubctl`), optional `organization_id`.
- Docs in `docs/source/reference/providers.rst`; offline unit tests in `tests/providers/`.

**Optional dependency** — `qhub-api`, via the `[kipu]` extra (`pip install perceval[kipu]`), imported
lazily. Base Perceval is unchanged without it; tests skip the Hub-backed cases cleanly when absent.

**Usage**
```python
import perceval as pcvl
import perceval.providers.kipu as kipu

session = kipu.Session(platform_name="quandela.sim.belenos", token="<your-PAT>")
processor = session.build_remote_processor()
# build a circuit, then sample as with any RemoteProcessor
```

Purely additive — no change to existing providers. `flake8` + `pytest tests/providers` pass with and without the extra.

### DIFF
--- a/.github/workflows/autotests.yml
+++ b/.github/workflows/autotests.yml
@@ -61,7 +61,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .
+        python -m pip install .[kipu]
         python -m pip install -r tests/requirements.txt
 
     - name: Lint with flake8
@@ -106,7 +106,6 @@ jobs:
         python -m pip install .
         python -m pip install -r tests/requirements.txt
         python -m pip uninstall -y -r .github/rendering_requirements.txt
-        python -m pip uninstall -y qhub-api
 
     - name: Test with pytest
       run: |
--- a/.github/workflows/autotests.yml
+++ b/.github/workflows/autotests.yml
@@ -106,6 +106,7 @@ jobs:
         python -m pip install .
         python -m pip install -r tests/requirements.txt
         python -m pip uninstall -y -r .github/rendering_requirements.txt
+        python -m pip uninstall -y qhub-api
 
     - name: Test with pytest
       run: |
--- a/perceval/providers/kipu/kipu_rpc_handler.py
+++ b/perceval/providers/kipu/kipu_rpc_handler.py
@@ -29,18 +29,19 @@
 
 from datetime import datetime
 
-from perceval.utils.logging import get_logger, channel
-
 _MISSING_QHUB_MSG = (
     "The Kipu Quantum Hub provider requires the 'qhub-api' package. "
     "Install it with: pip install perceval[kipu]"
 )
 
-_SUPPORTED_BACKENDS = ("quandela.sim.belenos", "quandela.qpu.belenos")
+_SIM_BELENOS = "quandela.sim.belenos"
+_QPU_BELENOS = "quandela.qpu.belenos"
+
+_SUPPORTED_BACKENDS = (_SIM_BELENOS, _QPU_BELENOS)
 
 _ALIASES = {
-    "sim:belenos": "quandela.sim.belenos",
-    "qpu:belenos": "quandela.qpu.belenos",
+    "sim:belenos": _SIM_BELENOS,
+    "qpu:belenos": _QPU_BELENOS,
 }
 
 _STATUS_MAP = {
@@ -107,14 +108,32 @@ def _import_qhub() -> dict:
         "client": HubQuantumClient,
         "credentials": DefaultCredentialsProvider,
         "input": {
-            "quandela.sim.belenos": CreateJobRequestInput_QuandelaSimBelenos,
-            "quandela.qpu.belenos": CreateJobRequestInput_QuandelaQpuBelenos,
+            _SIM_BELENOS: CreateJobRequestInput_QuandelaSimBelenos,
+            _QPU_BELENOS: CreateJobRequestInput_QuandelaQpuBelenos,
         },
         "params": {
-            "quandela.sim.belenos": CreateJobRequestInputParams_QuandelaSimBelenos,
-            "quandela.qpu.belenos": CreateJobRequestInputParams_QuandelaQpuBelenos,
+            _SIM_BELENOS: CreateJobRequestInputParams_QuandelaSimBelenos,
+            _QPU_BELENOS: CreateJobRequestInputParams_QuandelaQpuBelenos,
         },
     }
+
+
+def _build_httpx_client(proxies: dict | None):
+    """Build an httpx.Client routing through `proxies`, or None if none set.
+
+    Perceval proxies are requests-style ({scheme: url}); httpx wants per-scheme
+    transport mounts. httpx is imported lazily — it ships with qhub-api, not base
+    Perceval.
+    """
+    if not proxies:
+        return None
+    import httpx
+    mounts = {
+        f"{scheme}://": httpx.HTTPTransport(proxy=url)
+        for scheme, url in proxies.items()
+    }
+    # explicit timeout: injecting httpx_client bypasses qhub-api's 60s default
+    return httpx.Client(timeout=60, mounts=mounts)
 
 
 def _to_timestamp(value):
@@ -143,7 +162,9 @@ class KipuRPCHandler:
         resolved from the environment or the `qhubctl login` config file
     :param organization_id: optional Kipu organization id; when None your
         personal account is used
-    :param proxies: optional protocol->URL proxy mapping (stored for symmetry)
+    :param proxies: optional protocol->URL proxy mapping (requests-style, e.g.
+        ``{"https": "http://proxy:8080"}``); routed through the underlying httpx
+        client
     :param client: optional pre-built HubQuantumClient (used for testing)
     """
 
@@ -154,12 +175,6 @@ class KipuRPCHandler:
         self._token = token
         self._organization_id = organization_id
         self._proxies = proxies or {}
-        if self._proxies:
-            get_logger().warn(
-                "Kipu Quantum Hub provider does not support proxies; "
-                "the 'proxies' argument is ignored.",
-                channel.general,
-            )
         self._backend_id = _resolve_backend_id(platform_name)
         self._client = client if client is not None else self._build_client()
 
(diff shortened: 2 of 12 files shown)


.github/workflows/autotests.yml
```diff
@@ -106,6 +106,7 @@ jobs:
         python -m pip install .
         python -m pip install -r tests/requirements.txt
         python -m pip uninstall -y -r .github/rendering_requirements.txt
+        python -m pip uninstall -y qhub-api
```
Aubaert: Instead of uninstalling it here, I suggest to remove the package from the tests/requirements.txt, and do a `pip install .[kipu]` in the perceval-autotests workflow. This would remove the redundancy between the setup.py and the requirements.txt

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+# SOFTWARE.
+
+from datetime import datetime
+
+_MISSING_QHUB_MSG = (
+    "The Kipu Quantum Hub provider requires the 'qhub-api' package. "
+    "Install it with: pip install perceval[kipu]"
+)
+
+_SUPPORTED_BACKENDS = ("quandela.sim.belenos", "quandela.qpu.belenos")
```
Aubaert: Is there a way to get these (or at least confirm that the requested platform exists) from your API ? If so, can you use it to avoid hardcoding these ?

Edit: given the imports, this seems complicated to avoid hardcoding these

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+            f"Unknown Kipu platform '{platform_name}'. "
+            f"Supported backends: {', '.join(_SUPPORTED_BACKENDS)} "
+            f"(aliases: {', '.join(_ALIASES)})."
+        )
+    return backend_id
+
+
+def _to_perceval_status(hub_status: str | None) -> str:
+    """Map a Hub JobStatus value to a Perceval status string."""
+    if hub_status is None:
```
Aubaert: (Note that this case is included by the get(), but you can keep it like this).

Also, why calling str() if you already expect a str?

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+        fields (creation/start/duration). Progress and failure code are not
+        exposed by the Hub; the error message for a failed job comes from the
+        result payload via ``_fetch_error_message`` (surfaced as status_message).
+        """
+        status = _to_perceval_status(getattr(self._client.jobs.get_job_status(job_id), "status", None))
+        job = self._client.jobs.get_job(job_id)
+        status_message = self._fetch_error_message(job_id) if status == "error" else None
+        return {
+            "creation_datetime": _to_timestamp(getattr(job, "created_at", None)),
+            "duration": getattr(job, "runtime", None),
```
Aubaert: The default value should be a number (0.) , else, we could call int(None) or float(None)  (see `_retrieve_from_response` in remote_job.py)

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+        ``get_job_status`` returns the live status; ``get_job`` adds the timing
+        fields (creation/start/duration). Progress and failure code are not
+        exposed by the Hub; the error message for a failed job comes from the
+        result payload via ``_fetch_error_message`` (surfaced as status_message).
+        """
+        status = _to_perceval_status(getattr(self._client.jobs.get_job_status(job_id), "status", None))
+        job = self._client.jobs.get_job(job_id)
+        status_message = self._fetch_error_message(job_id) if status == "error" else None
+        return {
+            "creation_datetime": _to_timestamp(getattr(job, "created_at", None)),
```
Aubaert, line 249: Same

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+            "duration": getattr(job, "runtime", None),
+            "failure_code": None,
+            "last_intermediate_results": None,
+            "msg": "ok",
+            "progress": 1.0 if status == "completed" else 0.0,
+            "progress_message": None,
+            "start_time": _to_timestamp(getattr(job, "started_at", None)),
```
Aubaert, line 256: Same

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+        return message if isinstance(message, str) else None
+
+    def get_job_results(self, job_id: str) -> dict:
+        """Fetch Hub job results, normalized to Perceval's result envelope.
+
+        The Hub stores Quandela-native results already in Perceval's shape.
+        """
+        result = self._client.jobs.get_job_result(job_id)
+        return {
+            "duration": result.get("duration"),
```
Aubaert, line 283: From what I saw, only the "results" field is needed, but keeping the other ones should not cause any trouble

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+    :param client: optional pre-built HubQuantumClient (used for testing)
+    """
+
+    def __init__(self, platform_name, url=None, token=None, organization_id=None,
+                 proxies=None, client=None):
+        self._platform_name = platform_name
+        self._url = url
+        self._token = token
+        self._organization_id = organization_id
+        self._proxies = proxies or {}
```
Aubaert, line 177: I see proxies are not handled. You should log a warning if this field is not empty

perceval/providers/kipu/kipu_rpc_handler.py
```diff
@@ -0,0 +1,286 @@
...
+        input_cls = qhub["input"][self._backend_id]
+        params_cls = qhub["params"][self._backend_id]
+
+        job = self._client.jobs.create_job(
+            backend_id=self._backend_id,
+            shots=shots,
+            input=input_cls(value=inner),
+            input_format="PERCEVAL",
+            input_params=params_cls(
+                platform_name=payload.get("platform_name"),
```
Aubaert: Beware that the platform name that is received here is the name of the RemoteProcessor. If the user passes an alias, you will receive the alias here. I'm not sure how useful this parameter is though.

Aubaert (commented):
Hello,

Thank you for the addition. Globally, the code is nice and nicely integrated. I just had a few  small remarks before we can merge :)

.github/workflows/autotests.yml
wagnerse-kipu: Thanks a lot for your review. 

Done. perceval-autotests now runs pip install .[kipu], removed qhub-api from tests/requirements.txt, and dropped the redundant uninstall qhub-api step in the lite job

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Kept hardcoded. The provider intentionally exposes only the Quandela backends reachable via the Hub.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Simplified — dropped both the str() and the explicit None branch. _STATUS_MAP.get(hub_status, "unknown") covers None (not a key) on its own.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Fixed. Timing fields now default to a number (0.) instead of None, so the float()/int() coercion in _retrieve_from_response no longer logs an error on every poll when a field is absent.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Same fix as above.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: I added support for proxies.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Kept the extra fields. Only results is strictly required, but forwarding duration/job_id/results_type is harmless and keeps the envelope aligned with what RemoteJob reads.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Same fix as above.

perceval/providers/kipu/kipu_rpc_handler.py
wagnerse-kipu: Dropped it — it was redundant. The Hub already selects the target platform internally before sending the job creation request via your REST API. So the SDK doesn't need to send platform_name.

Aubaert (approved):

perceval/providers/kipu/kipu_rpc_handler.py
Aubaert: This is still to be done I think

Aubaert:
Hello, you can merge when you are ready

wagnerse-kipu:
Thanks for approving the PR. 

> Hello, you can merge when you are ready

Thanks! I don't have merge permissions on the repo (external contributor), so I can't merge from my side — could you merge when convenient?

PR #773, opened 2026-07-22
PCVL-1259: Remove deprecated features

cmarsone:


perceval/runtime/rpc_handler.py
```diff
@@ -126,11 +126,6 @@ def fetch_platform_details(self):
         endpoint = self.build_endpoint(_ENDPOINT_PLATFORM_DETAILS_NEW, quote_name, _PROCESSOR_SECTION)
         try:
             response = self.get_request(endpoint)
```
Aubaert, line 129: I think we can keep only this line and remove the whole try except

cmarsone: I got some tests failed for the remote tests if I remove the except section : Error while fetching platform details: 404 Client Error  Not Found for url: ...

Aubaert: Yes, we want to remove the whole _ENDPOINT_PLATFORM_DETAILS value, so you also need to remove the tests associated with it

PR #774, opened 2026-07-23
PCVL-1275 new workflow scaleway and kipu

Aubaert:


perceval/providers/kipu/__init__.py
```diff
@@ -1,3 +1,32 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
 # MIT License
```
Benoit-F-Q, line 30: Most `perceval/providers/kipu/*.py` and `tests/providers/kipu/*.py` have twin licence

perceval/runtime/communication_layer.py
```diff
@@ -26,51 +26,126 @@
...
+    def stop_session(self) -> None:
+        """May be used to stop a non-interrupted session. May do nothing on stateless implementations"""
+        pass
+
+    def delete_session(self) -> None:
+        """May be used to delete a non-interrupted session. May do nothing on stateless implementations"""
+        pass
+
+
+class RPCBasedCommunicationLayer(CommunicationLayer):
```
Benoit-F-Q, line 132: Do we expect non RPC-based communication layers?
If currently not, I think we should merge those classe

perceval/runtime/execution.py
```diff
@@ -202,15 +202,15 @@ def execute_sync(self, *args, allow_partial_results: bool = False, **kwargs) ->
             return self._results  # Problem here if we try to reuse a job with different args and kwargs
 
         self._status.start_run()
-        with self._computer.acquire():
```
Benoit-F-Q, line 205: this `acquire()` has been removed (not displaced).
Fine?

perceval/providers/kipu/kipu_communication_layer.py
```diff
@@ -0,0 +1,53 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+from perceval.runtime.communication_layer import RPCBasedCommunicationLayer
+
+from perceval.utils.logging import get_logger, channel
+
+from .kipu_rpc_handler import KipuRPCHandler
+
+
+class KipuCommunicationLayer(RPCBasedCommunicationLayer):
```
Benoit-F-Q, line 36: Dependency Injection guidelines promote
`RemoteComputer(CommunicationLayer(KipuRPCHandler(platform, token, organisation, url, proxies)))`
over `RemoteComputer(KipuCommunicationLayer(platform, token, organisation, url, proxies))` and agree it is painful.

Here we break this rule for user-friendness with a helper class, it could be more consistent to do it fully with `RemoteComputerKipu(platform, token, organisation, url, proxies)`.

Nonetheless, I do prefer the current version as it emphasizes that we're talking to a Quandela computer over Kipu network

Benoit-F-Q (approved):

perceval/providers/kipu/__init__.py
Aubaert: It's the same license, except for the copyright, which causes troubles with the pre-commit hook. I can explicitly tell the pre-commit to ignore this file

perceval/runtime/communication_layer.py
Aubaert: Using the RPC signatures has some constraints, especially having to look at the RPCBased layer to know what is expected in each field. Here, using this is mostly to simplify the changes and limit the risks, but you can also look at the ComputerProxy in the test_remote_computer file

perceval/runtime/execution.py
Aubaert: Yes, it was the decision taken with Scaleway. Users should include it in their script where they see fit

PR #776, opened 2026-07-28
PCVL-1273 adapt algorithms to new workflow

Aubaert:


tests/algorithm/test_processor_compatibility.py
```diff
@@ -0,0 +1,77 @@
+# MIT License
```
Benoit-F-Q, line 1: "retro" is french. "Backward compatibility"

perceval/algorithm/processor_compatibility.py
```diff
@@ -0,0 +1,101 @@
+# MIT License
```
Benoit-F-Q, line 1: "BackwardCompatibility" or more specific "ProcessorCompatibility"

perceval/algorithm/analyzer.py
```diff
@@ -28,36 +28,42 @@
...
                  input_states: list[BasicState] | dict[BasicState, str],
                  output_states=None,
                  mapping=None,
                  **kwargs):
         if mapping is None:
             mapping = {}
-        super().__init__(processor, **kwargs)
-        self._sampler = Sampler(processor, **kwargs)
+        super().__init__(computer, **kwargs)
+        self._experiment = experiment  # Make a copy ?
```
Benoit-F-Q: It depends whether the different methods assume that the underlying experiement remains the same.

As there is only one public method `compute()`, it seems a valid usecase to change the experiment while keeping the Analyzer

perceval/algorithm/abstract_algorithm.py
```diff
@@ -26,17 +26,24 @@
...
+from .retrocompatibility import computer_from_processor
 
 
 class AAlgorithm:
     _MAX_SHOTS_NAMED_PARAM = "max_shots_per_call"
 
-    def __init__(self, processor: AProcessor, **kwargs):
-        self._processor = processor
+    def __init__(self, computer: AbstractComputer, **kwargs):
+        # TODO: remove (deprecated since 1.3)
```
Benoit-F-Q, line 39: So we have to update notebooks too

perceval/algorithm/tomography/tomography_utils.py
```diff
@@ -46,20 +47,24 @@ def _compute_probs(tomography_experiment, prep_state_indices: list, meas_pauli_b
...
-    e = experiment_circuit_configurator(tomography_experiment._processor.experiment, prep_state_indices,
+    e = experiment_circuit_configurator(tomography_experiment._experiment, prep_state_indices,
                                         meas_pauli_basis_indices)
 
-    p = Processor(tomography_experiment._processor.backend, e)  # TODO: allow RemoteProcessor
     input_state = BasicState([1, 0] * tomography_experiment._nqubit)
-    p.with_input(input_state)
+    e.with_input(input_state)
+
+    parameters = tomography_experiment._computer.parameters | {"compute_physical_logical_perf": True}
```
Benoit-F-Q, line 56: `|` -> `or`?

perceval/algorithm/retrocompatibility.py
```diff
@@ -0,0 +1,101 @@
...
+from perceval.providers.kipu import KipuRPCHandler, KipuCommunicationLayer
+from perceval.providers.quandela import QuandelaCommunicationLayer
+from perceval.providers.quandela.rpc_handler import RPCHandler as QuandelaRPCHandler
+from perceval.providers.scaleway import ScalewayCommunicationLayer, RPCHandler as ScalewayRPCHandler
+
+
+# TODO: remove this file when removing the Processor support
+
+def computer_from_processor(processor: AProcessor) -> AbstractComputer:
+    # This method is a quick patch for places where a Processor is needed.
```
Benoit-F-Q: This method is a ~~quick~~ patch for places where a Processor is needed.
This baseline should be adapted to specific cases when needed

perceval/algorithm/analyzer.py
Aubaert: The problem is that `compute()` stores the results inside the Analyzer, and is called automatically when asking for the `distribution` property. Changing the experiment after calling compute will not change the results if someone asks the `distribution`, but it will if `compute()` has not been called before

perceval/algorithm/abstract_algorithm.py
Aubaert: The goal was to keep the Sampler as it is since we are going to drop it

perceval/algorithm/tomography/tomography_utils.py
Aubaert: For dictionaries, this is a union (a | b is equivalent to {**a, **b}, but more readable)

Benoit-F-Q: Not that readable through :)

Benoit-F-Q (approved):

PR #777, opened 2026-07-28
Pcvl 1255 rename global payload to cloud data

MegaiSa:


Aubaert:
Continued in https://github.com/Quandela/Perceval/pull/786

PR #782, opened 2026-08-06
PCVL-1274 job group for emt jobs

Aubaert:
Starts from the new serialization branch - Must be reviewed after it


Benoit-F-Q (approved):
Extra newline before `def _launch_wait_executions(...)`

PR #790, opened 2026-09-03
PCVL-1288 polish perceval new version

Aubaert:


perceval/components/detector.py
```diff
@@ -336,3 +337,26 @@ def check_heralds_detectors(heralds: dict[int, int] | None, detectors: list[IDet
...
+
+
+def update_detectors(detectors: list[IDetector], output_transmissions: list[float]) -> None:
+    """
+    Updates in-place the given detectors, by changing their efficiency. Only works for Detector type (not BSLayeredPPNR).
+    """
+    output_transmissions = [eff / max(output_transmissions) for eff in output_transmissions]
+
+    for i, (eff, d) in enumerate(zip(output_transmissions, detectors)):
+        if isinstance(d, Detector):
```
Benoit-F-Q: This pattern is flawed: if someone subclasses Detector, it will be lost here.

You have to either `d.clone_and_change_efficiency(eff)` or have a setter in `IDetector`.

Also note that `BSLayeredPPNR` won't behave as expected regarding this transmission factor

perceval/runtime/error_mitigation/abstract_mitigation.py
```diff
@@ -38,10 +38,21 @@
...
 
     APPLY_MIN_PHOTONS = True  # By default, avoid any accident at the cost of performance
     APPLY_LOGICAL_SELECTION = True
+    KNOWN_MITIGATIONS = []
+    _TAG: str
+
+    def __init_subclass__(cls, /, tag=None, **kwargs):
+        super().__init_subclass__(**kwargs)
+        tag = tag or cls.__name__
+        AMitigation.KNOWN_MITIGATIONS.append(tag)
```
Benoit-F-Q, line 53: raise on duplicate

perceval/runtime/abstract_computer.py
```diff
@@ -44,20 +44,20 @@
...
 
-class AbstractComputer(ABC):
+class AComputer(ABC):
     """
     A Computer, able to execute Computations, applying automatically error mitigations.
     It can handle custom parameters to change how the computation will be done.
     """
 
-    # Note: a computer must be hashable to be usable in ExecutionGroup,
+    # Note: a computer must be hashable to be usable in acquire(),
```
Benoit-F-Q, line 53: if not aquirable, is it usable at all?

If yes, is it still mandatory for ExecutionGroup?

perceval/runtime/local_computer.py
```diff
@@ -151,8 +154,20 @@ def _execute_command(self, computation: Computation, progress_callback: Progress
                 raise e
 
     def _execute_command_async(self, computation: Computation) -> _ThreadedGetter:
-        return _ThreadedGetter(self._execute_single, args=(computation,))
+        # Copy so external changes won't interfere
+        # TODO: The copy creates a problem with the reserve_resource() in case the lock is instance-bounded
+        # TODO: replace by deepcopy when available on all backends
```
Benoit-F-Q: Should be doable now

perceval/serialization/library/archive.py
```diff
@@ -122,8 +122,22 @@ def pre_record(self, children: Iterable[Any]):
         self.memo.extend( [ self.NoValue ] * (len(self.ids) - len(self.memo)) )
 
     # To storable object
-    def to_json(self):
-        raise NotImplementedError("JSON storage not implemented")
+    def to_json(self) -> dict:
+        """
+        :return: A dict representation of the archive, that can be converted to str using json.dumps
+        """
+        res = {"header": self.header,
```
Benoit-F-Q, line 129: I'm unsure whether this should be a different header for different export formats :/

tests/runtime/test_execution.py
```diff
@@ -179,3 +185,20 @@ def test_get_res_run_async():
 
     assert res_1st_call["results"] == res_2nd_call["results"]
     assert res_1st_call["global_perf"] == res_2nd_call["global_perf"]
+
+
+# TODO: restore this test
```
Benoit-F-Q: Why was it commented?

Should a `@pytest.disable()` be more appropriate?

perceval/runtime/abstract_computer.py
Aubaert: No, the only place where this is needed is the `acquire()` method because it is hashed to remove the duplicates, but `self.acquire()` can still be used if it's not

tests/runtime/test_execution.py
Aubaert: Because it was failing due to the computer copy not being deep enough. It should be good now

Benoit-F-Q (approved):

Benoit-F-Q (approved):

PR #791, opened 2026-09-14
RC Upgrades

Aubaert:


perceval/components/experiment.py
```diff
@@ -78,7 +78,7 @@ class Experiment:
     _no_copiable_attributes = { '_noise_changed_observers' }
 
     def __init__(self, m_circuit: int | ACircuit = None, noise: NoiseModel = None, name: str = "Experiment"):
-        self._input_state = None
+        self._input_state: FockState | AnnotatedFockState | SVDistribution | None = None
```
Benoit-F-Q, line 81: NoisyFS ?

perceval/runtime/remote_computer.py
```diff
@@ -185,27 +182,41 @@ def check_min_detected_photons_filter(experiment: Experiment) -> None:
...
 
         self.check_min_detected_photons_filter(experiment)
 
         constraints = self.specs.constraints
         if constraints:
-            input_state = experiment.input_state
-            n_heralds = sum(experiment.in_heralds.values())
-            n_photons = input_state.n + n_heralds
             # Checks on state
+            if not constraints.get("support_any_state_kind", False):
```
Benoit-F-Q: "if not (, false)" looks confusing.

Better define a list of accepted states here `[ FockState ]` or `[ FS SV ... ] `and then check if `input_state` is accepted.

Getting this list throught the cloud is more complicated + not needed yet, I'd advice stick the types here

perceval/components/experiment.py
Aubaert: SV and NoisyFS are transformed into SVD when given, so the internal value can't have these types

Benoit-F-Q (approved):

